Issue #134, opened 2016-08-02
parameter validation


stevendborrelli:
I don't know if this should be a macro, but I was bitten today in Ansible because interpolations resulted in an empty variable. Is it possible that we have some checks to make sure computed variables or outputs of a module meet certain conditions?

Ignore the crappy syntax, but something like: 

```
required {
       len(computed_var) > 0
       len(consul_servers_group) = len (control_servers_group)
       len(counsul_servers_group) >= 3
               "A minimum of 3 consul servers is required)
} 

warn {
        (consul_servers_group % 2) != 0 :
               "The number of consul servers should be an odd number"
}
```

sehqlr:
I'm going to remove this from the 0.3.0 project, in favor of tracking only the PR on that board

PR #219, opened 2016-09-06
Add param validation w/ type checks and predicates

sehqlr:
This feature adds the ability for users to constrain their inputs into a
converge system. The specification that this is based on is located in
the [wiki](https://github.com/asteris-llc/converge/wiki/Parameter-Types-and-Validation).

Things that work:

- you can specify an `int` as a value, and the validation loads in the
  DSL for checking the values, as specified

  The functions that are loaded into the templates are actually closures
  containing the value of the param with proper type during the Prepare
  stage, so we don't need to pipe in that value later on. This changes
  how the rules are wrapped within the validation logic, so this strays
  from the spec, but I think this will work better, and give the users
  more flexibility with rules that they write. The spec will need to be
  updated if everyone agrees.
- There is a table test that I wrote for this that works pretty well,
  and it should be easier to backfill with more test cases as we come up
  with more corner cases and types
- other tests that I didn't touch still passed (woo!)

Things that don't work yet:

- I did not make a specialized version of `len` for string types. We can
  include `len` from `text/template`. Because of the validator closures
  (as mentioned above), users can use `len . |` in their rules, which
  matches how golang devs operate with `text/template` anyhow. If the
  spec needs to be changed, it'll need to include documentation for the
  difference between the validator closures and the normal funcs
- the `description` field in a rule is not checked for, and is not
  included as a field in the struct. I forgot if that feature was going
  to be part of our "metadata for nodes" feature. Plus, I feel like it
  is a lower priority item for the demo, so I want to push it to last if
  need be
- loading a parameter with a rule stanza causes an error. I'll need to
  pair with someone to help work out what I'm doing wrong

TODOs:

- Backfill more test cases
- Add sample .hcl files to appropriate places

### DIFF
--- a/resource/param/preparer.go
+++ b/resource/param/preparer.go
@@ -16,16 +16,27 @@ package param
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
-	"math"
 	"strconv"
 	"text/template"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/asteris-llc/converge/load/registry"
-	"github.com/asteris-llc/converge/render/extensions"
 	"github.com/asteris-llc/converge/resource"
 )
+
+const (
+	PASS int = iota
+	WARN
+	FAIL
+	OTHER_ERR
+)
+
+// Rule for Type Validation
+//
+// Rule holds a predicate for parameter validation. It takes a text/template
+// fragment as input. The validation logic then wraps the expression around a
+// different template fragment, then evals it to determine the outcome
 
 // Preparer for params
 //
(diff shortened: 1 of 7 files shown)


arichardet:
Fixes #134

resource/param/preparer.go
```diff
@@ -15,40 +15,186 @@
...
-		return &Param{Value: val}, nil
+	value, present := render.Value()
+
+	if !present {
+		if p.Default == nil {
+			return nil, errors.Errorf("param is required")
+		}
+		def, err := render.Render("default", *p.Default)
+		if err != nil {
+			return nil, err //errors.Wrap(err, "rendering default failed")
```
ryane: can remove this comment if you decided not to wrap the error

samples/paramValidation.hcl
```diff
@@ -0,0 +1,41 @@
...
+  must = ["min 50", "max 512"]
+}
+
+task.query "sayconverge" {
+  interpreter = "/bin/bash"
+  query       = "echo -n converge"
+}
+
+param "converge" {
+  default = "{{lookup `task.query.sayconverge.checkstatus.stdout`}}"
```
ryane: hm, should this just be `status` instead `checkstatus`? Is there a reason where you might need one over the other? I just haven't seen this before

resource/param/preparer.go
```diff
@@ -15,40 +15,186 @@
...
+			return value, errors.Wrapf(err, `paramType is "int", but converting "%s" failed`, paramValue)
+		}
+		return value, nil
+
+	// this case is a nop, since string is the default type for parameters
+	case ParamTypeString:
+		return paramValue, nil
+	case ParamTypeInferred:
+		// try to infer the type from the value
+		// the recursion is make the code DRYer, but is this the best way?
```
ryane: I could see this getting unwieldy if we want to add more types for validation. I wonder if we just require that the `type` is specified when using validation in order to keep things simple. Or, perhaps better, maybe it's possible to build on the work in #321 to infer the type here.

samples/paramValidation.hcl
sehqlr: It should be status.

sehqlr:
After looking at the conflicts, I'm going to have to rework this PR to bring in the new changes. :le-sigh:

sehqlr:
This is going to take quite a lot of reworking, so I'm closing this for now

PR #221, opened 2016-09-07
file module

stevendborrelli:
Refactor file to be in a single module. Manages symlinks, hard links, directories, file content.

### DIFF
--- a/docs_source/content/resources/file.md
+++ b/docs_source/content/resources/file.md
@@ -12,30 +12,43 @@ Manages files, file content, directories, hard and soft links.
 ## Example
 
 ```hcl
-param "filename" {
-  default = "test.txt"
+param "dir" {
+  default = "/tmp/converge"
 }
 
-file "content" {
-  destination = "{{param `filename`}}"
-  mode        = 0777
+file "dir" {
+  destination = "{{param `dir`}}"
+  type        = "directory"
   state       = "present"
-  content     = "managed by converge"
+}
+
+file "hello" {
+  destination = "{{param `dir`}}/hello.txt"
+  mode        = "0750"
+  depends     = ["file.dir"]
 }
 
 file "symlink" {
-  destination = "symlink"
-  target      = "{{param `filename`}}"
+  destination = "{{param `dir`}}/symlink"
+  target      = "{{lookup `file.hello.Destination`}}"
+  state       = "present"
+  type        = "symlink"
+}
+
+file "deepdir" {
+  destination = "{{param `dir`}}/a/b/c/d/e"
+  type        = "directory"
   state       = "present"
 }
 
-file "directory" {
-  type        = directory
-  destination = "dir"
-  mode        = 07555
+# create parent dir using force 
+file "deepfile" {
+  destination = "{{param `dir`}}/a/b/c/d/e/f/g/deep.txt"
+  type        = "file"
+  state       = "present"
+  content     = "converge test"
+  force       = "true"
 }
-
-
 ```
 
 ## Parameters
--- a/docs_source/content/resources/file.md
+++ b/docs_source/content/resources/file.md
@@ -7,6 +7,7 @@ menu:
     parent: resources
 ---
 
+
 Manages files, file content, directories, hard and soft links.
 
 ## Example
@@ -69,8 +70,8 @@ file "deepfile" {
 
 - `group` (string)
 
-  The group account that owns the file. This account must exist on the system or an
-  error will be raised.
+  The group account that owns the file. This account must exist on the
+  system or an error will be raised.
 
   Default is the effective Group ID of the converge process. If unset, group ownership
   of existing files will not be changed.
--- a/docs_source/content/resources/file.md
+++ b/docs_source/content/resources/file.md
@@ -1,0 +1,93 @@
+---
+title: "File"
+slug: "file"
+date: "2016-10-04"
+menu:
+  main:
+    parent: resources
+---
+
+Manages files, file content, directories, hard and soft links.
+
+## Example
+
+```hcl
+param "filename" {
+  default = "test.txt"
+}
+
+file "content" {
+  destination = "{{param `filename`}}"
+  mode        = 0777
+  state       = "present"
+  content     = "managed by converge"
+}
+
+file "symlink" {
+  destination = "symlink"
+  target      = "{{param `filename`}}"
+  state       = "present"
+}
+
+file "directory" {
+  type        = directory
+  destination = "dir"
+  mode        = 07555
+}
+
+
+```
+
+## Parameters
+
+- `destination` (required string)
+
+  Destination specifies which file will be modified by this resource.
+
+- `force` (boolean)
+
+  Force the change. This is requried for:
+
+  - changing the file type
+  - changing an existing hard link
+  - creating a file where the parent directory does not exist
+
+  Default is `false`
+
+- `group` (string)
+
+  The group account that owns the file. This account must exist on the system or an
+  error will be raised.
+
+  Default is the effective Group ID of the converge process. If unset, group ownership
+  of existing files will not be changed.
+
+- `mode` (required base 8 optional uint32)
+
+  Mode is the mode of the file, specified in octal.
+
+  Default is the `0750`. If unset, permissions on existin files will not be changed.
+
+- `state` (string)
+
+  Whether the file should be `present` or `absent`. If `absent` is requested,
+  the file will be removed from the system.
+
+  Default is `present`
+
+- `target` (string)
+
+  Required for hard and soft links, this is the target file that `destination`
+  is linked to. `destination -> target`
+
+  This file must exist for a hard link, or an error will be raised.
+
+  Set `force = true` to change the target of a hard link.
+
+- `user` (string)
+
+  The user account that owns the file. This account must exist on the system or an
+  error will be raised.
+
+  Default is the effective User ID of the converge process. If unset, the ownership
+  of existing files will not be changed.
--- a/docs_source/content/resources/file.md
+++ b/docs_source/content/resources/file.md
@@ -41,7 +41,7 @@ file "deepdir" {
   state       = "present"
 }
 
-# create parent dir using force 
+# create parent dir using force
 file "deepfile" {
   destination = "{{param `dir`}}/a/b/c/d/e/f/g/deep.txt"
   type        = "file"
--- a/docs_source/content/resources/file.md
+++ b/docs_source/content/resources/file.md
@@ -7,7 +7,6 @@ menu:
     parent: resources
 ---
 
-
 Manages files, file content, directories, hard and soft links.
 
 ## Example
--- a/resource/file/file.go
+++ b/resource/file/file.go
@@ -41,23 +41,27 @@ var validLinkTypes = []string{"hardlink", "symlink"}
 
 // File contains information for managing files
 type File struct {
-	Destination   string
-	State         string
-	Type          string
-	Target        string
-	Force         bool        //force replacement of symlinks, etc
-	Mode          string      //requested permissions from Prepare
-	FileMode      os.FileMode //calculated permissions
-	UserInfo      *user.User
-	GroupInfo     *user.Group
-	Content       []byte
+	Destination string
+	State       string
+	Type        string
+	Target      string
+	Force       bool        //force replacement of symlinks, etc
+	Mode        string      //requested permissions from Prepare
+	FileMode    os.FileMode //calculated permissions
+	UserInfo    *user.User
+	GroupInfo   *user.Group
+	Content     []byte
+
 	action        string //create, delete, modify
 	modifyContent bool   // does content need to be changed
 
+	*resource.Status
+	renderer resource.Renderer
 }
 
 // Apply  changes to file resources
-func (f *File) Apply() error {
+func (f *File) Apply(r resource.Renderer) (resource.TaskStatus, error) {
+
 	var err error
 	switch f.action {
 	case "delete":
@@ -68,56 +72,59 @@ func (f *File) Apply() error {
 		err = f.Modify()
 	}
 	if err != nil {
-		return errors.Wrapf(err, "%s on %s failed: %s", f.action, f.Destination)
+		return f.Status, errors.Wrapf(err, "%s on %s failed: %s", f.action, f.Destination)
 	}
-	return err
+	return f.Status, err
 }
 
 // Check File settings
-func (f *File) Check() (resource.TaskStatus, error) {
-	status := &resource.Status{Status: f.Destination}
-	status.Output = append(status.Output, fmt.Sprintf("%s (%s)", f.Destination, f.Type))
+func (f *File) Check(r resource.Renderer) (resource.TaskStatus, error) {
+	f.renderer = r
+	f.Status = &resource.Status{}
+	f.Status.Output = append(f.Status.Output, fmt.Sprintf("%s (%s)", f.Destination, f.Type))
+
 	var actual *File
+
 	// Get information about the current file
 	stat, err := os.Lstat(f.Destination) //link aware
 
 	if os.IsNotExist(err) { //file not found
 		switch f.State {
 		case "absent": // if "absent" is set and the file doesn't exist, return with no changes
-			status.WillChange = false
-			status.WarningLevel = resource.StatusNoChange
-			return status, nil
+			return f, nil
 		case "present": //file doesn't exist, we need to create it
-			actual = &File{Destination: "<file does not exist>", State: "absent", UserInfo: &user.User{}, GroupInfo: &user.Group{}}
-			status.WillChange = true
-			status.WarningLevel = resource.StatusWillChange
-			f.diffFile(actual, status)
+			actual = &File{Destination: "<file does not exist>",
+				State:     "absent",
+				UserInfo:  &user.User{},
+				GroupInfo: &user.Group{},
+				Status:    &resource.Status{}}
+			f.diffFile(actual)
 			f.action = "create"
 		}
 	} else { //file exists
-		actual = &File{Destination: f.Destination, State: "present"}
+		actual = &File{
+			Destination: f.Destination,
+			State:       "present"}
 		switch f.State {
 		case "absent": //remove file
-			status.WillChange = true
-			status.AddDifference("destination", actual.Destination, "<removed>", "")
-			status.AddDifference("state", actual.State, f.State, "")
+			f.Status.WillChange = true
+			f.Status.WarningLevel = resource.StatusWillChange
 			f.action = "delete"
 		case "present": //modify file
 			err = GetFileInfo(actual, stat)
 			if err != nil {
-				status.WarningLevel = resource.StatusFatal
-				return status, fmt.Errorf("unable to get file info for %s: %s", f.Destination, err)
+				f.Status.WarningLevel = resource.StatusFatal
+				return f, fmt.Errorf("unable to get file info for %s: %s", f.Destination, err)
 			}
 			actual.Content, _ = Content(actual.Destination)
-			f.diffFile(actual, status)
-			if status.WillChange {
+			f.diffFile(actual)
+			if f.Status.WillChange {
 				f.action = "modify"
 			}
-
 		}
 	}
 
-	return status, nil
+	return f, nil
 }
 
 // Validate runs checks against a File resource
@@ -219,7 +226,7 @@ func (f *File) validateUser() error {
 	if f.UserInfo.Username != "" {
 		u, err := user.Lookup(f.UserInfo.Username)
 		if err != nil {
-			return fmt.Errorf("unable to get user information for username %s:", f.UserInfo.Username, err)
+			return fmt.Errorf("unable to get user information for username %s: %s", f.UserInfo.Username, err)
 		}
 		f.UserInfo.Username = u.Username
 	}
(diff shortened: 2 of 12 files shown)


resource/file/file_test.go
```diff
@@ -0,0 +1,254 @@
...
+		for _, tt := range tests {
+			name := filepath.Base(tt.F.Destination)
+			t.Run("check-"+name, func(t *testing.T) {
+				status, err := tt.F.Check(fakerenderer.New())
+				assert.Equal(t, tt.Err, err)
+				assert.Equal(t, tt.HasChanges, status.HasChanges(), fmt.Sprintf("Differences %+v", status.Diffs()))
+			})
+
+			t.Run("apply-"+name, func(t *testing.T) {
+				_, err = tt.F.Apply()
```
arichardet: You should probably test that status is what you expect.

resource/file/file_test.go
```diff
@@ -0,0 +1,254 @@
...
+		{File{Destination: filepath.Join(tmpDir, "file-missing"), State: "absent"}, nil, false},
+		{File{Destination: filepath.Join(tmpDir, "file-content"), Content: []byte("converge/content test\n"), State: "absent"}, nil, true},
+		{File{Destination: filepath.Join(tmpDir, "dir-A"), Mode: mode(0700), State: "absent", Type: "directory"}, nil, true},
+		{File{Destination: filepath.Join(tmpDir, "dir-missing"), State: "absent"}, nil, false},
+		{File{Destination: filepath.Join(tmpDir, "dir-A/dir-B"), State: "absent", Type: "directory"}, nil, false},       //parent dir removed first
+		{File{Destination: filepath.Join(tmpDir, "dir-C"), State: "absent", Type: "directory", Force: true}, nil, true}, //remove child dirs
+		{File{Destination: filepath.Join(tmpDir, "file-symlink"), State: "absent", Target: filepath.Join(tmpDir, "file-content"), Type: "symlink"}, nil, true},
+		{File{Destination: filepath.Join(tmpDir, "file-hardlink"), State: "absent", Target: filepath.Join(tmpDir, "file-test"), Type: "hardlink"}, nil, true},
+	}
+
```
arichardet: It looks like these test cases check successes; we probably need some tests for the different ways failure can occur.

resource/file/util.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+//given two groups, decide which one to use
+//returns true if a change is required
+func desiredGroup(f, actual *user.Group) (groupInfo *user.Group, changed bool, err error) {
+	if f == nil || f.Name == "" {
+		if actual == nil || actual.Name == "" { // if neither is set, use the effective gid of the process
+			groupInfo, err = user.LookupGroupId(strconv.Itoa(os.Getegid()))
+			if err != nil {
+				return &user.Group{}, true, errors.Wrapf(err, "unable to set default group")
```
stevendborrelli: Yes. Fixed.

resource/file/util.go
```diff
@@ -0,0 +1,168 @@
...
+			return userInfo, true, err
+		}
+		if actual.Username != "" {
+			return &user.User{Username: actual.Username, Uid: actual.Uid}, true, err
+		}
+	}
+
+	userInfo, err = user.Lookup(f.Username)
+	if err != nil {
+		return userInfo, true, errors.Wrapf(err, "unable to get user information for username %s:", f.Username)
```
stevendborrelli: fixed

resource/file/file.go
```diff
@@ -0,0 +1,558 @@
...
+		switch f.Force {
+		case true:
+			// if the file type changes, delete and recreate
+			err = f.Delete()
+			if err != nil {
+				return errors.Wrapf(err, "apply: unable to recreate")
+			}
+			f.action = ActionCreate
+		default:
+			return errors.New("apply: please set force=true to change file")
```
stevendborrelli: I've added it to diffFile called during check

resource/file/util.go
```diff
@@ -0,0 +1,180 @@
...
+//given two users, decide which one to use
+//returns true if a change is required
+func desiredUser(f, actual *user.User) (userInfo *user.User, changed bool, err error) {
+	if f == nil || f.Username == "" {
+		if actual == nil || actual.Username == "" { // if neither is set, use the effective uid of the process
+			userInfo, err = user.LookupId(strconv.Itoa(os.Geteuid()))
+			if err != nil {
+				return &user.User{}, false, errors.Wrapf(err, "unable to set default username %s", f.Name)
+			}
+			return userInfo, true, err
```
arichardet: Probably better to return nil instead of err to be more clear...it would be good to change throughout the code.

resource/file/util.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		if actual.Username != "" {
+			return &user.User{Username: actual.Username, Uid: actual.Uid}, false, err
+		}
+	}
+
+	userInfo, err = user.Lookup(f.Username)
+	if err != nil {
+		return userInfo, true, errors.Wrapf(err, "unable to get user information")
```
arichardet: Should this be false?

resource/file/util.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+//given two users, decide which one to use
+//returns true if a change is required
+func desiredUser(f, actual *user.User) (userInfo *user.User, changed bool, err error) {
+	if f == nil || f.Username == "" {
+		if actual == nil || actual.Username == "" { // if neither is set, use the effective uid of the process
+			userInfo, err = user.LookupId(strconv.Itoa(os.Geteuid()))
+			if err != nil {
+				return &user.User{}, false, errors.Wrapf(err, "unable to set default username %s", f.Name)
```
arichardet: should probably return nil instead of user.User

resource/file/util.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+//given two groups, decide which one to use
+//returns true if a change is required
+func desiredGroup(f, actual *user.Group) (groupInfo *user.Group, changed bool, err error) {
+	if f == nil || f.Name == "" {
+		if actual == nil || actual.Name == "" { // if neither is set, use the effective gid of the process
+			groupInfo, err = user.LookupGroupId(strconv.Itoa(os.Getegid()))
+			if err != nil {
+				return &user.Group{}, false, errors.Wrapf(err, "unable to set default group")
```
arichardet: should probably return nil instead of user.Group

resource/file/util.go
```diff
@@ -0,0 +1,180 @@
...
+			return groupInfo, true, nil
+		}
+		if actual.Name != "" { //if we didn't request a group, use the file's information
+			return &user.Group{Name: actual.Name, Gid: actual.Gid}, false, nil
+		}
+	}
+
+	groupInfo, err = user.LookupGroup(f.Name)
+	if err != nil {
+		return groupInfo, true, errors.Wrapf(err, "unable to get information")
```
arichardet: should changed = false since it failed?

resource/file/util.go
stevendborrelli: Changed

resource/file/util.go
stevendborrelli: Yes, I'll change them.

resource/file/util.go
stevendborrelli: yes, I have moved it to nil

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	return &File{
+		UserInfo:  &user.User{},
+		GroupInfo: &user.Group{},
+	}
+}
+
+// Apply changes to file resources
+func (f *File) Apply() (resource.TaskStatus, error) {
+
+	status, err := f.diff(New())
```
stevendborrelli: I've changed the function, it is just `diff()` now.

docs_source/content/resources/file.md
```diff
@@ -0,0 +1,106 @@
...
+
+file "hello" {
+  destination = "{{param `dir`}}/hello.txt"
+  mode        = "0750"
+  depends     = ["file.dir"]
+}
+
+file "symlink" {
+  destination = "{{param `dir`}}/symlink"
+  target      = "{{lookup `file.hello.Destination`}}"
```
BrianHicks: should use lower-case lookups

docs_source/content/resources/file.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Parameters
+
+- `destination` (required string)
+
+  Destination specifies which file will be modified by this resource.
+
+- `force` (boolean)
+
+  Force the change. This is requried for:
```
BrianHicks: "required"

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	"os"
+	"os/user"
+	"path/filepath"
+	"strconv"
+
+	"github.com/asteris-llc/converge/resource"
+	"github.com/pkg/errors"
+)
+
+const defaultPermissions = os.FileMode(0750)
```
BrianHicks: Why executable by default?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+	// ActionModify deletes a File resource
+	ActionModify
+
+	// ActionDelete deletes a File resource
+	ActionDelete
+)
+
+func (a Action) String() string {
+	return string(a)
```
BrianHicks: this will return a number, which won't mean anything to the user. I'd suggest:

```go
switch a {
  case 0:
    return "none"

  case 1:
    return "create"

  // et cetera
}
```

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+// State is the desired state of the file
+type State string
+
+const (
+	// StateAbsent indicates the file should be absent
+	StateAbsent State = "absent"
+
+	// StateUndefined indicates an undefined state
+	StateUndefined State = ""
```
BrianHicks: this case should be handled in the preparer, since it looks like you're just substituting another value everywhere.

resource/file/validate.go
```diff
@@ -0,0 +1,158 @@
...
+	return err
+}
+
+// Validate the state or set default value
+func (f *File) validateState() error {
+	switch f.State {
+	case StateUndefined:
+		f.State = DefaultState
+		return nil
+	default:
```
BrianHicks: +1

resource/file/validate.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+// Validate the type or set default value
+func (f *File) validateType() error {
+	switch f.Type {
+	case "": //use default if not set
+		f.Type = DefaultType
+		return nil
+	default:
```
BrianHicks: +1

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+	// StatePresent indicates the file should be present
+	StatePresent State = "present"
+
+	// DefaultState is the default state
+	DefaultState State = StatePresent
+)
+
+// ValidStates indicates supported states that can be defined
+var ValidStates = []State{StateAbsent, StatePresent}
```
BrianHicks: remove in favor of `valid_values`

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	TypeDirectory Type = "directory"
+
+	// TypeFile is a regular file
+	TypeFile Type = "file"
+
+	// TypeLink is a hardlink
+	TypeLink Type = "hardlink"
+
+	// TypeNone is an undefined type
+	TypeNone Type = ""
```
BrianHicks: same here as with `StateUndefined`

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+var (
+	// ValidFileTypes are valid file types
+	ValidFileTypes = []Type{TypeDirectory, TypeFile}
+
+	// ValidLinkTypes are valid link types
+	ValidLinkTypes = []Type{TypeLink, TypeSymlink}
+
+	// AllTypes are the current FileTypes that are supported
+	AllTypes = append(ValidFileTypes, ValidLinkTypes...)
```
BrianHicks: 1. looks like `ValidFileTypes` and `ValidLinkTypes` aren't being used anywhere but in this definition
2. Use `valid_values` instead, please

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	"strconv"
+
+	"github.com/asteris-llc/converge/resource"
+	"github.com/pkg/errors"
+)
+
+const defaultPermissions = os.FileMode(0750)
+
+// Action is the calculated action
+type Action uint8
```
rebeccaskinner: It doesn't look like this is being used outside of the package- can it be made package private?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	// ActionDelete deletes a File resource
+	ActionDelete
+)
+
+func (a Action) String() string {
+	return string(a)
+}
+
+// State is the desired state of the file
+type State string
```
rebeccaskinner: Private for this too?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+// ValidStates indicates supported states that can be defined
+var ValidStates = []State{StateAbsent, StatePresent}
+
+func (s State) String() string {
+	return string(s)
+}
+
+// Type is the type of file
+type Type string
```
rebeccaskinner: Also private?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+}
+
+// File contains information for managing files
+type File struct {
+	Destination string
+	State       State
+	Type        Type
+	Target      string
+	Force       bool    //force replacement of file
+	Mode        *uint32 //requested permissions from Prepare
```
BrianHicks: this probably should be a non-pointer FileMode type by the time it comes from Prepare

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	UserInfo    *user.User
+	GroupInfo   *user.Group
+	Content     []byte
+
+	action Action //create, delete, modify
```
BrianHicks: nit: our comments have a space after the comment token. `// comment`  rather than `//comment`

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	status.AddMessage(fmt.Sprintf("%s (%s)", f.Destination, f.Type))
+	if err != nil || status.Level == resource.StatusFatal {
+		return status, errors.Wrap(err, "check")
+	}
+
+	return status, err
+}
+
+// Compare desired state with an actual file (if present)
+func (f *File) diff() (*resource.Status, error) {
```
rebeccaskinner: If we return a `*resource.Status` instead of embedded it in anything then we won't be able to lookup the fields on it.  It might be better to create a return type like:

```
type FileStatus struct {
  Destination string
  Content string
  // anything else that should be `lookup`-able
  *resource.Status
}
```

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+		status.Level = resource.StatusFatal
+		status.AddMessage(err.Error())
+	}
+	return status, err
+}
+
+// Check File settings
+func (f *File) Check(r resource.Renderer) (resource.TaskStatus, error) {
+
+	f.renderer = r
```
BrianHicks: this is the only place you're using this. Maybe drop the field?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+	action Action //create, delete, modify
+
+	renderer resource.Renderer
+}
+
+// New returns a File with UserInfo and GroupInfo fields allocated
+func New() *File {
+	return &File{
+		UserInfo:  &user.User{},
```
BrianHicks: how is this better than a nil pointer?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+			err = f.diffFile(actual, status)
+			if err != nil {
+				status.Level = resource.StatusFatal
+				return status, errors.Wrapf(err, "diff")
+			}
+			return status, nil
+		default:
+			return status, fmt.Errorf("unknown state %s", f.State)
+		}
+	} else {
```
rebeccaskinner: It looks like we'll always return from `if os.IsNotExit(err)` so we could move this logic out of `else`

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	return status, err
+}
+
+// Check File settings
+func (f *File) Check(r resource.Renderer) (resource.TaskStatus, error) {
+
+	f.renderer = r
+
+	status, err := f.diff()
+	status.AddMessage(fmt.Sprintf("%s (%s)", f.Destination, f.Type))
```
BrianHicks: this message isn't actionable. What does it convey to the user seeing this for the first time?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+
+	switch f.action {
+	case ActionDelete:
+		err = f.Delete()
+	case ActionCreate, ActionModify:
+		err = f.Modify(status)
+	}
+	if err != nil {
+		status.Level = resource.StatusFatal
+		status.AddMessage(err.Error())
```
BrianHicks: these two can be together: `status.SetError(err)`

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	}
+
+	return status, err
+}
+
+// Compare desired state with an actual file (if present)
+func (f *File) diff() (*resource.Status, error) {
+
+	status := &resource.Status{}
+	var stat os.FileInfo
```
BrianHicks: why is this declared up here? What benefit does it give you above `stat, err := ...` below?

resource/file/file.go
rebeccaskinner: There is a lot of deeply nested logic here with a number of exit points.  Is there any way to factor some of the logic out into functions so that `diff()` itself is just managing the branching logic and the functions are managing the other logic?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	// AllTypes are the current FileTypes that are supported
+	AllTypes = append(ValidFileTypes, ValidLinkTypes...)
+)
+
+func (t Type) String() string {
+	return string(t)
+}
+
+// File contains information for managing files
+type File struct {
```
BrianHicks: File should be embedding a status so we can look up the fields later

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+			actual := &File{
+				Destination: "<file does not exist>",
+				State:       StateAbsent,
+				Type:        f.Type,
+				UserInfo:    &user.User{},
+				GroupInfo:   &user.Group{},
+			}
+			err = f.diffFile(actual, status)
+			if err != nil {
+				status.Level = resource.StatusFatal
```
BrianHicks: returning an error will set this status for you. Alternatively, you can use `SetError` and return it that way.

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+		case StateAbsent: // if "absent" is set and the file doesn't exist, return with no changes
+			return status, nil
+
+		case StatePresent: // file doesn't exist, we need to create it
+			f.action = ActionCreate
+			actual := &File{
+				Destination: "<file does not exist>",
+				State:       StateAbsent,
+				Type:        f.Type,
+				UserInfo:    &user.User{},
```
BrianHicks: how are these more useful than a nil pointer?

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+				GroupInfo:   &user.Group{},
+			}
+			err = f.diffFile(actual, status)
+			if err != nil {
+				status.Level = resource.StatusFatal
+				return status, errors.Wrapf(err, "diff")
```
BrianHicks: this error will be displayed to the user. "diff: errmsg" doesn't strike me as being particularly useful. Also, you can use `errors.Wrap` here since you don't have any formatting directives in the message.

resource/file/file.go
BrianHicks: all the branches in the conditional above return something. You can pop this out of the `else` block. That's what you'll usually see in Go code.

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+			return status, fmt.Errorf("unknown state %s", f.State)
+		}
+	} else {
+		actual := &File{
+			Destination: f.Destination,
+			State:       StatePresent,
+		}
+		switch f.State {
+		case StateAbsent: //file exists -> absent
+			status.Level = resource.StatusWillChange
```
BrianHicks: use `RaiseLevel` or equiv

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+			if err != nil {
+				return errors.Wrapf(err, "modify: unable to recreate")
+			}
+			f.action = ActionCreate
+		default:
+			return errors.New("modify: please set force=true to change file")
+		}
+	}
+
+	switch f.Type {
```
rebeccaskinner: This feels like another opportunity to factor the implementation into helper functions called inside of the `case` statements.

resource/file/file.go
```diff
@@ -0,0 +1,566 @@
...
+	Target      string
+	Force       bool    //force replacement of file
+	Mode        *uint32 //requested permissions from Prepare
+	UserInfo    *user.User
+	GroupInfo   *user.Group
+	Content     []byte
+
+	action Action //create, delete, modify
+
+	renderer resource.Renderer
```
rebeccaskinner: Is this being used for anything?

BrianHicks:
Fixes #327

PR #235, opened 2016-09-12
Feature/systemd unit

Dacode45:
Added a systemd.unit task 

``` hcl
systemd.unit "test" {
  name   = "foo.service"
  active = "true"
  state  = "enabled"

  content = <<EOF
[Unit]
Description=Foo hello world printer
[Service]
ExecStart=/bin/bash -c "while true; do /bin/echo HELLO WORLD; sleep 5; done;"

[Install]
WantedBy=multi-user.target
EOF
}
```

**Note: the content parameter will not create the unit file. It should already exist on disk**


Dacode45:
Are people having dependency issues with this branch? Could you let me know the error message?

resource/systemd/states.go
```diff
@@ -0,0 +1,186 @@
...
+	ASFailed       ActiveState = "failed"
+	ASActivating   ActiveState = "activating"
+	ASDeactivating ActiveState = "deactivating"
+)
+
+func (state ActiveState) Equal(state2 ActiveState) bool {
+	if state == state2 {
+		return true
+	}
+	state = ActiveState(strings.Replace(string(state), "\"", "", -1))
```
BrianHicks: are you trying to remove quotes here? Seems like you ought to remove them higher up the stack so you don't ever get an invalid value in a state. If you *do* have to do it here though: `strings.Trim(string(state), "\"") == strings.Trim(string(state), "\"")` works just fine.

Oh, and it might read better if `state2` was called `other`.

resource/systemd/states.go
```diff
@@ -0,0 +1,186 @@
...
+		Name:  "UnitFileState",
+		Value: godbus.MakeVariant(string(ufs)),
+	}
+}
+
+func (state UnitFileState) Equal(state2 UnitFileState) bool {
+	if state == state2 {
+		return true
+	}
+	state = UnitFileState(strings.Replace(string(state), "\"", "", -1))
```
BrianHicks: same comment here about de-quoting as soon as possible and not waiting.

resource/systemd/status.go
```diff
@@ -0,0 +1,48 @@
...
+		a.Differences = b.Differences
+	} else {
+		for key, value := range b.Differences {
+			a.Differences[key] = value
+		}
+	}
+	a.Output = append(a.Output, b.Output...)
+	if b.WarningLevel > a.WarningLevel {
+		a.WarningLevel = b.WarningLevel
+		a.Status = b.Status
```
BrianHicks: warning: status is going to be gone when you rebase. Just watch out for that.

resource/systemd/status.go
```diff
@@ -0,0 +1,48 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package systemd
+
+import "github.com/asteris-llc/converge/resource"
+
+func AppendStatus(a, b *resource.Status) *resource.Status {
```
BrianHicks: We really shouldn't be doing this kind of thing. Why are you creating more than one Status?

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Preparer for Systemd Unit
+type Preparer struct {
+	// The name of this unit as in "foo.service"
+	Name string `hcl:"name"`
+
+	// Active determines whether this unit should be active or inactive
+	Active string `hcl:"active"`
+
+	/* State is the UnitFileState of the unit.
```
BrianHicks: Could you add some whitespace lines in this comment? It's nearly unreadable as-is.

BrianHicks: Aslo you can add markdown formatting in these and they'll be translated into the docs :)

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,129 @@
...
+	// suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns",
+	// "us" (or "µs"), "ms", "s", "m", "h".
+	Timeout string `hcl:"timeout" doc_type:"duration string"`
+
+	// The location of the original uit file.
+	Destination string `hcl:destination"`
+}
+
+func (p *Preparer) Prepare(render resource.Renderer) (resource.Task, error) {
+	name, err := render.Render("name", p.Name)
```
BrianHicks: The need to do this is going to go away with #321

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+// TODO Enable Parallelization of this Unit
+type Unit struct {
+	// The name of this unit as in "foo.service"
+	Name string // (either just file names or full absolute paths if the unit files are residing outside the usual unit search paths
+
+	// Active determines whether this unit should be active or inactive
+	Active bool
+
+	/*
+		UnitFileState encodes the install state of the unit file of FragmentPath.
```
BrianHicks: why is this indented like this? Also, you don't need to repeat the doc block here, Preparer is sufficient for fields you've already documented.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+	Timeout time.Duration // how long to wait for the load state to resolve
+}
+
+// Check if all the properties for the unit are correct
+// 1. Checks if the unit is currently loading, if so waits Default 5 seconds
+// 2. Checks if the unit is in the active state
+// 3. Check if the unit is in the unit file state
+func (t *Unit) Check(r resource.Renderer) (resource.TaskStatus, error) {
+	systemd.ApplyDaemonReload()
+	/*////////////////////////////////////////////////////
```
BrianHicks: please don't decorate comments like this.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+	First thing to do is to check if the Name given is outside
+	the normal search path of systemd. If so it should be linked
+	*/ ////////////////////////////////////////////////////
+	dir, unitName := filepath.Split(t.Name)
+	var shouldBeLinked = false
```
BrianHicks: choose `var shouldBeLinked bool` or `shouldBeLinked := false`. They do the same thing.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+	ctx, cancel = context.WithTimeout(context.Background(), t.Timeout)
+	defer cancel()
+
+	// Waits for the daemon to finish loading.
+	// After check for unit linking
+	err = systemd.WaitToLoad(ctx, dbusConn, unitName)
+
+	if err != nil {
+		// Unit doesn't exist should be linked
+		if err.Error() == "LoadError: \"[\\\"org.freedesktop.DBus.Error.FileNotFound\\\", \\\"No such file or directory\\\"]\"" {
```
BrianHicks: using backtick syntax here will keep you from having to escape this as much.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+					return &resource.Status{
+						Status:       err.Error(),
+						WarningLevel: resource.StatusFatal,
+						Output:       []string{err.Error()},
+					}, err
+				}
+				statusMsg := fmt.Sprintf("unit %q does not exist, will be linked", unitName)
+				diffMsg := fmt.Sprintf("unit %q does not exist", unitName)
+				diff := resource.TextDiff{Default: diffMsg, Values: [2]string{diffMsg, fmt.Sprintf("unit %q is linked", unitName)}}
+				return &resource.Status{
```
BrianHicks: status should be embedded in the Task now. Check the module authors' guide at converge.aster.is

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+
+	statusMsg := fmt.Sprintf("unit %q is in unknown state, %s", unitName, state)
+	return &resource.Status{
+		Status:       statusMsg,
+		WarningLevel: resource.StatusFatal,
+		Output:       []string{statusMsg},
+	}, fmt.Errorf(statusMsg)
+}
+
+func (t *Unit) StartOrStop(dbusConn *dbus.Conn) (*resource.Status, error) {
```
BrianHicks: why are these two functions together? Shouldn't this have a separate stop and start function?

Also all these utility methods should be private.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,543 @@
...
+	statusMsg := fmt.Sprintf("unit %q is %s", unitName, activeMsg)
+	return &resource.Status{
+		Status: statusMsg,
+		Output: []string{statusMsg},
+	}, nil
+}
+
+var validUnitFileStates = systemd.UnitFileStates{systemd.UFSEnabled, systemd.UFSEnabledRuntime, systemd.UFSLinked, systemd.UFSLinkedRuntime, systemd.UFSDisabled}
+
+func (t *Unit) Validate() error {
```
BrianHicks: this validation belongs inline in Preparer.Prepare

resource/systemd/wait.go
```diff
@@ -0,0 +1,88 @@
...
+	"time"
+
+	"github.com/asteris-llc/converge/resource"
+	"github.com/coreos/go-systemd/dbus"
+)
+
+const DefaultTimeout = time.Second * 5
+
+func WaitToLoad(ctx context.Context, conn *dbus.Conn, unit string) error {
+	// first checck if there was an error loading ocnfiguration
```
BrianHicks: spelling here

resource/systemd/wait.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func WaitForLoadedState(ctx context.Context, conn *dbus.Conn, unit string) error {
+	err := conn.Subscribe()
+	if err != nil {
+		return err
+	}
+	// @ Reviewer. function may complete faster by putting this function in a
+	// looping gorutine constatnly pooling for the state instead of subscribing to state changes
+	// ever arbitrary time period. this would take more cpu cycles.
```
BrianHicks: 100ms is plenty quick enough to detect a change. If we need it to be faster we can lower the check value.

BrianHicks: That said, is there no way to subscribe to a feed of changes in that API?

samples/systemd.hcl
```diff
@@ -0,0 +1,5 @@
+systemd.unit "test" {
+  name   = "/tmp/foo8.service"
```
BrianHicks: could we have a more descriptive name here? Also, this seems to be an incomplete sample since the service file is undefined.

BrianHicks:
Also next time you push you should get codeclimate violations. Could you fix those before the next review, please?

BrianHicks:
And run `make vendor-clean` to get rid of extraneous files as well, please. Sorry, I keep remembering things!

resource/systemd/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+package systemd
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+func MultiErrorAppend(errs ...error) error {
```
Dacode45: MultiError has a bug where if you append two nil errors you get an error instead of nil. This will
 rectify that.

resource/systemd/pool.go
```diff
@@ -0,0 +1,57 @@
...
+		case connectionPool <- c.Connection:
+			c.Connection = nil
+		default:
+			c.Connection.Close()
+			c.Connection = nil
+		}
+	}
+}
+
+var connectionPool chan *dbus.Conn
```
Dacode45: sync.Pool wouldn't call `.Close()` when there are too many dbus connections.

BrianHicks:
Why are you appending nil in the first place?

BrianHicks:
In that case, what’s the mechanism for returning connections? Looks 
like it’s not being called in this file.

resource/systemd/unit/unit.go
Dacode45: It's less boilerplate code to have them be the same function. Starting Or Stopping is just a one line difference.

BrianHicks:
Reducing boilerplate is fine, but this looks like it’s actually a 
toggle function. Undefined behavior ahoy!

resource/systemd/daemon_reload.go
```diff
@@ -5,6 +5,9 @@ import (
 	"github.com/coreos/go-systemd/dbus"
 )
 
+//CheckDaemonReload checks whether the systemd daemon needs to be reloaded.
+//it does not check if a unit is failed and needs to be reset though.
+//NOTE : should I implement checkResetFailed?
```
BrianHicks: the semantics there probably should go in. How would you go about that?

resource/systemd/states.go
```diff
@@ -66,13 +66,13 @@ const (
...
+func (state ActiveState) Equal(other ActiveState) bool {
+	if state == other {
 		return true
 	}
-	state = ActiveState(strings.Replace(string(state), "\"", "", -1))
-	state2 = ActiveState(strings.Replace(string(state2), "\"", "", -1))
-	return state == state2
+	stateStr := strings.Trim(string(state), "\"")
+	otherStr := strings.Trim(string(other), "\"")
+	return stateStr == otherStr
```
BrianHicks: what even is the situation where you need to dequote these?

resource/systemd/errors.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
+// MultiErrorAppend appends errors together into one skipping nil errors.
+// If all errors passed in are nil then it also returns nil
 func MultiErrorAppend(errs ...error) error {
```
BrianHicks: again, you shouldn't need this because you shouldn't be appending nil errors. Can you please fix that case and remove this code?

resource/systemd/unit/preparer.go
```diff
@@ -25,42 +25,33 @@ import (
...
+	/* state is the `UnitFileState` of the unit.
+	`UnitFileState` encodes the install state of the unit file of FragmentPath.
+	It currently knows the following states: `enabled`, `enabled-runtime`, `linked`,
+	`linked-runtime`, `masked`, `masked-runtime`, `static`, `bad`, `disabled`, `invalid`.
+
+	Of these states only  `enabled`, `enabled-runtime`, `linked`,
+	`linked-runtime`, `masked`, `masked-runtime`, `static`, `disabled`, and
+	`invalid` are permitted in the preparer
+
+	See https://godoc.org/github.com/coreos/go-systemd/dbus for more info
```
BrianHicks: should be a markdown link.

resource/systemd/unit/preparer.go
```diff
@@ -25,42 +25,33 @@ import (
...
-	possibly replacing already queued jobs that conflict with this. If "fail" the
-	call will start the unit and its dependencies, but will fail if this would
-	change an already queued job. If "isolate" the call will start the unit in
-	question and terminate all units that aren't dependencies of it. If
-	"ignore-dependencies" it will start a unit but ignore all its dependencies.
-	If "ignore-requirements" it will start a unit but only ignore the requirement
-	dependencies. It is not recommended to make use of the latter two options.
-	Returns the newly created job object.
+
+	See https://godoc.org/github.com/coreos/go-systemd/dbus for more info
```
BrianHicks: same here.

resource/systemd/unit/preparer.go
BrianHicks: But we should also be explaining how to use our implementation here as well instead of punting the reader off to another page. Could you summarize?

resource/systemd/unit/preparer.go
```diff
@@ -121,7 +112,22 @@ func (p *Preparer) Prepare(render resource.Renderer) (resource.Task, error) {
 		StartMode:     systemd.StartMode(sm),
 		Timeout:       t,
 	}
-	return unit, unit.Validate()
+	return unit, Validate(unit)
+}
+
+var validUnitFileStates = systemd.UnitFileStates{systemd.UFSEnabled, systemd.UFSEnabledRuntime, systemd.UFSLinked, systemd.UFSLinkedRuntime, systemd.UFSMasked, systemd.UFSMaskedRuntime, systemd.UFSDisabled}
+
+func Validate(t *Unit) error {
```
BrianHicks: this should go *inline* into Prepare. It'll make more sense why we're doing that after #321

resource/systemd/unit/unit.go
```diff
@@ -124,42 +86,42 @@ func (t *Unit) Check(r resource.Renderer) (resource.TaskStatus, error) {
 
 	if err != nil {
 		// Unit doesn't exist should be linked
-		if err.Error() == "LoadError: \"[\\\"org.freedesktop.DBus.Error.FileNotFound\\\", \\\"No such file or directory\\\"]\"" {
+		if err.Error() == loadError {
 			if shouldBeLinked {
+				fmt.Println("Here--------------\n\n")
```
BrianHicks: debug statement, please remove

resource/systemd/unit/unit.go
```diff
@@ -124,42 +86,42 @@ func (t *Unit) Check(r resource.Renderer) (resource.TaskStatus, error) {
...
 		}
-		status := &resource.Status{
-			WarningLevel: resource.StatusFatal,
-			Status:       err.Error(),
+		t.TaskStatus = &resource.Status{
+			Level:  resource.StatusFatal,
+			Output: []string{err.Error()},
 		}
-		return status, err
+		return t.TaskStatus, err
```
BrianHicks: the return value here should be `t`. Since you're embedding the TaskStatus, we can do lookups on the task fields as well.

resource/systemd/unit/unit_test.go
```diff
@@ -49,11 +49,11 @@ func TestInactiveToActiveUnit(t *testing.T) {
 	defer svc.Remove()
 
 	foo := unit.Unit{Name: svc.Name, Active: true, UnitFileState: "enabled", StartMode: "replace"}
-	foo.Apply(&fr)
+	foo.Apply()
 	status, err := foo.Check(&fr)
 	assert.NoError(t, err)
 	assert.Equal(t, resource.StatusNoChange, status.StatusCode())
-	assert.Equal(t, fmt.Sprintf("property \"UnitFileState\" of unit %q is \"enabled\", expected one of [\"enabled, static\"]", svc.Name), status.Value())
+	assert.Equal(t, fmt.Sprintf("property \"UnitFileState\" of unit %q is \"enabled\", expected one of [\"enabled, static\"]", svc.Name), status.Messages()[0])
```
BrianHicks: this will cause a segfault if messages is empty. Check out assert.Contains.

resource/systemd/wait.go
```diff
@@ -52,7 +52,7 @@ func WaitToLoad(ctx context.Context, conn *dbus.Conn, unit string) error {
 		return err
 	}
 	//If loading wait until it becomes stable
```
BrianHicks: missing a space before this comment.

resource/systemd/states.go
Dacode45: Occasionally property values are "\"<property-value>\"". sometimes they are not. I have to take both into account.

resource/systemd/daemon_reload.go
Dacode45: In Check: 
1. Check if unit is in failed loading state. 
2. Return to the user 

In Apply
1. Check if unit is in a failed loading state.
2. Reset Failed
3. Reload Daemon
4. Wait to Load
5. If unit is no longer failed, continue.

BrianHicks:
If you’re dealing with untrusted or flaky input, you need to clean it 
as close to the source as possible. Don’t defer that work. Can you 
point me to where the values are coming from? Could this be a 
deserialization issue?

BrianHicks:
Does doing those things match up with the expectations the user is 
setting in the input files? If so, add them. If not, we still need to 
deal with failures. A `LevelCantChange` may be appropriate, but if we 
can fix the problem it’s even better.

resource/systemd/daemon_reload.go
Dacode45: I'll be heading out soon, but if this functionality is necessary, I'll implement it.

resource/systemd/states.go
Dacode45: This happens in the CheckProperty function when I get prop.Value.Value(). From my test It seems prop.Value.Value() returns "\"<property>\"", but prop.Value.String() returns "<property>". It also happens occasionally in the WaitForLoadedState function, where I get `unitStatus` in the select statement.

BrianHicks:
Then you should definitely clean it up at that point, not here. Does 
that module documentation say anything about that?

Dacode45:
I believe I have resolved all issues. 
- Input from the dbus is now clean
- Task now resets the failed state of units when necessary.

resource/systemd/daemon_operations.go
```diff
@@ -0,0 +1,53 @@
+package systemd
```
BrianHicks: needs our license header

resource/systemd/unit/unit.go
```diff
@@ -265,7 +265,7 @@ func (t *Unit) Apply() (resource.TaskStatus, error) {
 		}
 		return t, err
 	}
-	state := systemd.UnitFileState(prop.Value.String())
+	state := systemd.UnitFileState(prop.Value.Value().(string))
 
 	//////////////////////////
```
BrianHicks: more fancy comments. No, please

resource/systemd/daemon_operations.go
Dacode45: Done

resource/systemd/unit/unit.go
Dacode45: Done

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"time"
+
+	"github.com/asteris-llc/converge/load/registry"
+	"github.com/asteris-llc/converge/resource"
+	"github.com/asteris-llc/converge/resource/systemd"
+)
+
+// Preparer for Systemd Unit
+type Preparer struct {
```
ryane: preparer can be updated to use the new features introduced in #321.

ryane: see https://github.com/asteris-llc/converge/blob/55bc29870152341c4eed90acfaca6dab37328454/docs_source/content/resource-authors-guide.md#preparer

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+	// The name of this unit as in `foo.service`
+	// (either just file names or full absolute paths if the unit files are
+	// residing outside the usual unit search paths
+	Name string `hcl:"name"`
```
ryane: should this (or any of the other preparer fields) be marked as required?

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Active determines whether this unit should be in an active or inactive state
+	Active string `hcl:"active"`
```
ryane: you can just make this a bool now

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+	units, units that have no `[Install]` section, should be disabled with `masked` or
+	`masked-runtime`. Converge will not throw a warning if you want your unit to be
+	`enabled`, when it is infact `static`, and technically cannot be `enabled`,
+	since the unit will be "enabled" either way.
+	Likewise it will not throw a warning if you want your unit disabled, but it is
+	actually masked.
+
+	See [dbus](https://godoc.org/github.com/coreos/go-systemd/dbus) for more info
+	*/
+	UnitFileState string `hcl:"state"`
```
ryane: use `valid_values`

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+
+	/* Mode for the call to StartUnit()
+	StartUnit() enqeues a start job, and possibly depending jobs.
+	Takes the unit to activate, plus a mode string. The mode needs to be one of
+	replace, fail, isolate, ignore-dependencies, ignore-requirements.
+
+	See [dbus](https://godoc.org/github.com/coreos/go-systemd/dbus) for more info
+	*/
+	StartMode string `hcl:"mode"`
```
ryane: `valid_values`

resource/systemd/unit/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+	// "us" (or "µs"), "ms", "s", "m", "h".
+	Timeout string `hcl:"timeout" doc_type:"duration string"`
+
+	// The location of the original unit file.
+	Destination string `hcl:"destination"`
+}
+
+//Prepare returns a new systemd.unit task
+func (p *Preparer) Prepare(render resource.Renderer) (resource.Task, error) {
+	name, err := render.Render("name", p.Name)
```
ryane: a lot of this boilerplate can be removed

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,501 @@
...
+				Level: resource.StatusNoChange,
+				Output: []string{
+					statusMsg,
+					fmt.Sprintf("symlink for %q was %q. link between %q and %q", unitName, changes[0].Type, changes[0].Filename, changes[0].Destination),
+				},
+			}
+			return t, nil
+		}
+		// if hasEnaablementInfo is false then the unit will become static
+		hasEnaablementInfo, changes, err := dbusConn.EnableUnitFiles([]string{unitName}, t.UnitFileState.IsRuntimeState(), true)
```
ryane: hasEnablementInfo?

resource/systemd/check.go
```diff
@@ -0,0 +1,74 @@
...
+
+package systemd
+
+import (
+	"github.com/asteris-llc/converge/resource"
+	"github.com/coreos/go-systemd/dbus"
+)
+
+// CheckUnitIsActive determines if the unit property "ActiveState" is active
+func CheckUnitIsActive(conn *dbus.Conn, unit string) (*resource.Status, error) {
```
ryane: do all of these functions need to be exported?

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,501 @@
...
+
+	// Next check that the ActiveState property matches
+	activeState := systemd.ASActive
+	if !t.Active {
+		activeState = systemd.ASInactive
+	}
+	validActiveStates := []*dbus.Property{
+		systemd.PropActiveState(activeState),
+	}
+	asStatus, asErr := systemd.CheckProperty(dbusConn, unitName, "ActiveState", validActiveStates)
```
ryane: is there a reason you don't use the functions in check.go for this?

resource/systemd/check.go
ryane: are they being used anywhere?

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,501 @@
...
+// Unit represents a systemd units
+// NOTE: This unit does not parallelize with itself
+type Unit struct {
+	Name          string
+	Active        bool
+	UnitFileState systemd.UnitFileState
+	StartMode     systemd.StartMode
+	Timeout       time.Duration
+
+	resource.TaskStatus
```
rebeccaskinner: It's recommended that you embed an actual implementation of the status field, e.g. `*resource.Status`, that will make the fields accessible from `lookup`.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,501 @@
...
+		t.Timeout = systemd.DefaultTimeout
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), t.Timeout)
+	defer cancel()
+
+	// Waits for the daemon to finish loading.
+	// After check for unit linking
+	err = systemd.WaitToLoad(ctx, dbusConn, unitName)
+
+	if err != nil {
```
rebeccaskinner: It might make sense to break this out into it's own function since execution will end inside of the block.

resource/systemd/unit/unit.go
```diff
@@ -0,0 +1,501 @@
...
+
+	statusMsg := fmt.Errorf("unit %q is in unknown state, %s", unitName, state)
+	t.TaskStatus = &resource.Status{
+		Level:  resource.StatusFatal,
+		Output: []string{statusMsg.Error()},
+	}
+	return t, statusMsg
+}
+
+func (t *Unit) startOrStop(dbusConn *dbus.Conn) (resource.TaskStatus, error) {
```
rebeccaskinner: It feels like stopping and starting could be separate functions here.

resource/systemd/unit/unit_test.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// IsRoot checks if the current user is Root
+func IsRoot() bool {
+	currentUser, _ := user.Current()
+	return currentUser.Uid == "0"
+}
+
+// HasSystemd checks if dbus is available
+func HasSystemd() bool {
```
rebeccaskinner: It's a bit concerning that none of the unit tests can run on a system without systemd available.  Have you considered possibly seeing if there are any tests that you could run without having systemd available.  Usually unit tests should be able to run on every architecture that we plan on supporting, and integration tests should run on the relevant platforms.  This way if someone changes a higher level API that might break the module they'll be able to tell even if they're developing on a system without systemd.

It might be worth looking into either mocking the systemd library or breaking apart some of the functionality into smaller units so that you can at least have some test coverage for systems without systemd available.

resource/systemd/status.go
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package systemd
+
+import "github.com/asteris-llc/converge/resource"
+
+// AppendStatus combines two *resource.Status
+func AppendStatus(a, b *resource.Status) *resource.Status {
```
Dacode45: I'd like to. Should I do that in this PR or make a new PR for that?

stevendborrelli:
This fails on `make xcompile` on OS X:
```
3 errors occurred:
--> freebsd/386 error: exit status 2
Stderr: # github.com/asteris-llc/converge/vendor/github.com/godbus/dbus
vendor/github.com/godbus/dbus/transport_unix.go:51: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)
vendor/github.com/godbus/dbus/transport_unix.go:57: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)

--> freebsd/arm error: exit status 2
Stderr: # github.com/asteris-llc/converge/vendor/github.com/godbus/dbus
vendor/github.com/godbus/dbus/transport_unix.go:51: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)
vendor/github.com/godbus/dbus/transport_unix.go:57: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)

--> freebsd/amd64 error: exit status 2
Stderr: # github.com/asteris-llc/converge/vendor/github.com/godbus/dbus
vendor/github.com/godbus/dbus/transport_unix.go:51: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)
vendor/github.com/godbus/dbus/transport_unix.go:57: cannot use t (type *unixTransport) as type transport in return argument:
	*unixTransport does not implement transport (missing SendNullByte method)

make: *** [xcompile] Error 1
```

Dacode45:
Addressed Issues

rebeccaskinner:
In the future it's probably better to rebase onto master instead of merging it into your branch that way all of the changes from master don't get mixed into your change set during review.

rebeccaskinner:
Could you add a few tests to `unit_test.go` that will work on non-linux systems?  There's 0% test coverage on OSX and it would be nice to be able to catch regressions locally.  I can help you figure out how to approach that if you'd like.

stevendborrelli:
After a successful run, the systemd unit reports changes even though there are no changes to be made: (tasted on a centos 7 system)

```
root/systemd.unit.helloUnit:
 Messages:
 Has Changes: no
 Changes:
  hello.service:ActiveState:shouldbe="active":   "property \"ActiveState\" is \"active\""   => "property \"ActiveState\" should be one of [\"active\"]"
  hello.service:UnitFileState:shouldbe="linked": "property \"UnitFileState\" is \"linked\"" => "property \"UnitFileState\" should be one of [\"linked\"]"

```
```
status -l hello  
hello.service - Foo hello world
   Loaded: loaded (/tmp/hello.service; linked)
   Active: active (running) since Wed 2016-10-12 20:49:23 BST; 2min 56s ago
 Main PID: 16472 (bash)
   CGroup: /system.slice/hello.service
           ├─16472 /bin/bash -c while true; do /bin/echo HELLO WORLD; sleep 5; done;
           └─16627 sleep 5
```

Dacode45:
Unfortunately I couldn't figure out a way to mock systemd for testing, I have vetted this resource with systemd units on my system. I think this is ready to merge.

BrianHicks:
A lot has changed in the module implementation interface since you opened this, so I'm going to close this for now until we get a chance to rework it. Thanks @Dacode45!

PR #262, opened 2016-09-17
Make notes match sample code

tomduckering:
sample code places a file. The notes say it installs a package with YUM.

Correcting the notes.


BrianHicks:
Good catch, thanks @tomduckering!

Issue #254, opened 2016-09-14
Node is sometimes unresolvable, depending on ordering within a template


rebeccaskinner:
This is broken...
```hcl
task.query "should-install-go" {
  interpreter = "/bin/bash"
  query = "[[ '{{param `goversion`}}' == '{{lookup `task.query.go-installed-version.checkstatus.stdout`}}']]"
}
```

but this isnt...
```hcl
task.query "should-install-go" {
  interpreter = "/bin/bash"
  query = "[[ '{{lookup `task.query.go-installed-version.checkstatus.stdout`}}' == '{{param `goversion`}}' ]]"
}
```

in 

```hcl
task.query "lookup-gopath" {
  interpreter = "/bin/bash"
  query = " if [[ -z $GOPATH ]]; then echo -n '{{param `defaultGopath`}}'; else echo -n ${GOPATH}; fi "
}

task.query "go-installed-version" {
  interpreter = "/bin/bash"
  query = "go version | awk '{print $3}'"
}

param "goversion" {
  default = "1.7.1"
}

task.query "should-install-go" {
  interpreter = "/bin/bash"
  query = "[[ '{{lookup `task.query.go-installed-version.checkstatus.stdout`}}' == '{{param `goversion`}}' ]]"
}

param "defaultGopath" {
  default = "${HOME}/go"
}

param "gopath" {
  default = "{{lookup `task.query.lookup-gopath.status.stdout`}}"
}

param "platform" {
  default = "darwin-amd64"
}

param "gopkg" {
  default = "https://storage.googleapis.com/golang/{{param `packagename`}}"
}

param "packagename" {
  default = "go{{param `goversion`}}.{{param `platform`}}.pkg"
}

task.query "tempdir" {
  interpreter = "/bin/bash"
  query = "echo -n $(mktemp -d)"
}

param "downloadpath" {
  default = "{{lookup `task.query.tempdir.Status.Stdout`}}/{{param `packagename`}}"
}

task "download go" {
  interpreter = "/bin/bash"
  check = "[[ 1 == {{lookup `task.query.should-install-go.status.exitstatus`}} && -f {{param `downloadpath`}} ]]"
  apply = "curl {{param `gopkg`}} > {{param `downloadpath`}}"
}

```

rebeccaskinner:
Minimal reproduction:

```hcl
param "bar" {
  default = "bar"
}
task.query "foo" {
  query = "echo foo"
}
task.query "should-install-go" {
[...]

PR #265, opened 2016-09-19
Use `MakeLanguage` for dependecy resolution to ensure that no

rebeccaskinner:
unimplemented calls result in early template execution termination.

Fixes issue #254

### DIFF
--- a/render/extensions/extensions.go
+++ b/render/extensions/extensions.go
@@ -80,7 +80,7 @@ func MinimalLanguage() *LanguageExtension {
 
 // DefaultLanguage provides a default language extension.  It creates default
 // implementations of context-free and non-dependency-generating functions
-// (e.g. split) and provides a unimplemented function for functions that must bes
+// (e.g. split) and provides a unimplemented function for functions that must be
 // supplied with context or which may register dependencies.
 func DefaultLanguage() *LanguageExtension {
 	language := MakeLanguage()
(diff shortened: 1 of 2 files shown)


rebeccaskinner:
Build failure was caused by a closely related bug.  Basic fix is trivial but I am taking a little time to consider whether or not there's an approach to prevent this class of bug in the future.

render/extensions/extensions.go
```diff
@@ -64,9 +64,23 @@ func MakeLanguage() *LanguageExtension {
...
+	language.On("platform", newStub(&platform.Platform{}))
+	language.On("param", newStub(""))
+	language.On(RefFuncName, newStub(""))
+	return language
+}
+
 // DefaultLanguage provides a default language extension.  It creates default
 // implementations of context-free and non-dependency-generating functions
-// (e.g. split) and provides a unimplemented function for functions that must be
+// (e.g. split) and provides a unimplemented function for functions that must bes
```
BrianHicks: BEEEEES

render/extensions/extensions.go
```diff
@@ -158,6 +172,12 @@ func StubTemplateFunc(...string) (string, error) {
 	return "", nil
 }
 
+func newStub(returnVal interface{}) func(...string) (interface{}, error) {
```
BrianHicks: so this is basically just an `always`, right? Just trying to make sure I understand the semantics here.

PR #267, opened 2016-09-19
example: docker swarm mode

ryane:
provisions a multi-node docker swarm mode cluster on AWS using terraform and converge


examples/docker-swarm-mode/README.md
```diff
@@ -0,0 +1,41 @@
+# converge-docker-swarm
+
+A converge sample that sets up a multi-node [Docker Swarm Mode](https://docs.docker.com/engine/swarm/) cluster on AWS.
+
+## Usage
+
+### Terraform (AWS)
+
+You must have a version of https://github.com/ChrisAubuchon/terraform-provisioner-converge built and configured as a plugin for terraform:
```
BrianHicks: linkify?

.gitignore
```diff
@@ -38,3 +38,6 @@ build
 docs_source/extract
 docs_source/resources.mk
 docs_source/public
+
+# examples
+terraform.tfstate*
```
BrianHicks: won't this only ignore tfstates in the root? `**/terraform.tfstate*`

examples/docker-swarm-mode/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```shell
+$ cat ~/.terraformrc
+provisioners {
+  converge = "/path/to/terraform-provisioner-converge"
+}%
```
BrianHicks: looks like the lack of trailing newline crept in here

examples/docker-swarm-mode/README.md
```diff
@@ -0,0 +1,41 @@
...
+[centos@ip-10-0-1-239 ~]$ docker node ls
+ID                           HOSTNAME       STATUS  AVAILABILITY  MANAGER STATUS
+4omysuebc73ixajlymum3rrvp    ip-10-0-1-40   Ready   Active
+8dkplbxr7xluoyf0s91jumrr9 *  ip-10-0-1-239  Ready   Active        Leader
+b1k1r5hx9xe4ek34azbdx0ghe    ip-10-0-1-186  Ready   Active
+```
+
+## Notes
+
+* Uses the terraform graph engine to coordinate the timing of joining worker nodes to the cluster. The manager and worker nodes are created and provisioned in parallel. Then, using a null_resource, we join the workers to the cluster after the manager is initialized.
```
BrianHicks: This feels like maybe it should be its own paragraph? And also be a complete sentence to start.

examples/docker-swarm-mode/README.md
```diff
@@ -0,0 +1,41 @@
...
+* Does not currently support multiple manager nodes.
```
BrianHicks: Complete sentence here too

examples/docker-swarm-mode/converge-hcl/main.hcl
```diff
@@ -0,0 +1,11 @@
+param "docker-group-user-name" {}
```
BrianHicks: is this the name of the group or user?

ryane:
naming: 

* I used docker-swarm-mode to distinguish from the older docker swarm. they are two different things: https://docs.docker.com/engine/swarm/
* I acutally renamed to `converge-hcl` from `converge` b/c of the .gitignore setting. The current entry causes anything named `converge` to be ignored throughout the whole project. I will try and change the gitignore to ignore `converge` only in the root.

.gitignore
ryane: nope, it ignores them all. This behavior is the reason I renamed `converge` to `converge-hcl`

examples/docker-swarm-mode/converge-hcl/main.hcl
ryane: user. The value of this param is the name of the user that should be added to the docker group. 

As a side note, I originally wanted to be able to name it `docker-group-users` and provide a list of user names but I don't think we can do that (yet), correct?

BrianHicks: user-name feels sufficient.

ryane: now I remember why I had named it this way. In the docker.hcl module it is just `user-name` but I decided to namespace it in the top-level module. Since we don't actually have a naming conflict right now, I will go back and rename it to just `user-name`.

ryane:
I think I covered everything in the latest commits

BrianHicks:
LGTM!

PR #274, opened 2016-09-20
Make the logs quieter

### LINKED ISSUE #248: logging is a little too chatty
We log the stages right now, but it's not really actionable information. That should move to a debug level and only output operator-actionable information by default.

BrianHicks:
Fixes #248

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -76,7 +76,7 @@ real happens.`,
 
 		verifyModules := viper.GetBool("verify-modules")
 		if !verifyModules {
-			clog.Warn("skipping module verfication")
+			clog.Warn("skipping module verification")
 		}
 
 		// execute files
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -76,14 +76,14 @@ real happens.`,
 
 		verifyModules := viper.GetBool("verify-modules")
 		if !verifyModules {
-			clog.Warn("skipping module verfiction")
+			clog.Warn("skipping module verfication")
 		}
 
 		// execute files
 		for _, fname := range args {
 			flog := clog.WithField("file", fname)
 
-			flog.Info("applying")
+			flog.Debug("applying")
 
 			stream, err := client.Apply(
 				ctx,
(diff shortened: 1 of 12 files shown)


cmd/apply.go
```diff
@@ -76,14 +76,14 @@ real happens.`,
 
 		verifyModules := viper.GetBool("verify-modules")
 		if !verifyModules {
-			clog.Warn("skipping module verfiction")
+			clog.Warn("skipping module verfication")
```
rebeccaskinner: did you mean: "verification"?

cmd/healthcheck.go
```diff
@@ -53,12 +53,12 @@ not display healthy checks.`,
 
 		verifyModules := viper.GetBool("verify-modules")
 		if !verifyModules {
-			log.WithField("component", "client").Warn("skipping module verfiction")
+			log.WithField("component", "client").Warn("skipping module verfiaction")
```
rebeccaskinner: "verification"

rpc/executor.go
```diff
@@ -106,7 +106,7 @@ func (e *executor) Plan(in *pb.LoadRequest, stream pb.Executor_PlanServer) error
 	logger = logger.WithField("function", "executor.Plan")
 
 	if err := e.auth.authorize(ctx); err != nil {
-		logger.WithError(err).Info("authorization failed")
+		logger.WithError(err).Warning("authorization failed")
```
rebeccaskinner: Is an authorization failure a `Warning` or an `Error`?

BrianHicks: On the server, it's a warning since a user failing authentication should be raised for audit purposes but it's a normal occurrence. It's (still) an error on the client side.

BrianHicks:
Made the spelling fixes, thanks for the catch @rebeccaskinner. I'll rebase after your next review :)

rpc/executor.go
rebeccaskinner: Ah, that makes sense.

BrianHicks:
The lint errors here are due to sample output, which we maybe shouldn't change. They're in code blocks anyway, so I'm not sure what it's complaining about.

Issue #275, opened 2016-09-20
Allow adding and deleting a group without specifying gid


arichardet:

BrianHicks:
I'm gonna put this in 0.2.0 since it's a bug

BrianHicks:
Didn't close with #283. Weird. Fixed, thanks @arichardet!

PR #283, opened 2016-09-20
Fix/user group - Allow adding/deleting without gid

arichardet:
- Allow adding and deleting a group without supplying `gid`
- Update docs to indicate `state` default if not provided

### DIFF
--- a/resource/group/group.go
+++ b/resource/group/group.go
@@ -42,10 +42,10 @@ type Group struct {
 
 // SystemUtils provides system utilities for group
 type SystemUtils interface {
-	AddGroup(string, string) error
-	DelGroup(string) error
-	LookupGroup(string) (*user.Group, error)
-	LookupGroupID(string) (*user.Group, error)
+	AddGroup(groupName, groupID string) error
+	DelGroup(groupName string) error
+	LookupGroup(groupName string) (*user.Group, error)
+	LookupGroupID(groupID string) (*user.Group, error)
 }
 
 // ErrUnsupported is used when a system is not supported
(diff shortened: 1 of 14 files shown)


arichardet:
Fixes #275

resource/group/group.go
```diff
@@ -42,7 +42,7 @@ type Group struct {
 
 // SystemUtils provides system utilities for group
 type SystemUtils interface {
-	AddGroup(string, string) error
+	AddGroup(string, map[string]string) error
```
BrianHicks: could you add a comment to explain why this needs two arguments? Or just name the arguments?

resource/group/group_test.go
```diff
@@ -73,400 +73,597 @@ func init() {
...
-	m.On("LookupGroup", g.Name).Return(grp1, nil)
-	m.On("LookupGroupID", g.GID).Return(grp2, nil)
-	m.On("DelGroup", g.Name).Return(nil)
-	status, err := g.Apply()
-
-	m.AssertNotCalled(t, "DelGroup", g.Name)
-	assert.EqualError(t, err, fmt.Sprintf("will not attempt delete: group %s with gid %s", g.Name, g.GID))
-	assert.Equal(t, resource.StatusCantChange, status.StatusCode())
-}
+		assert.Equal(t, g.GID, options["gid"])
```
BrianHicks: why is this a map of strings instead of a real struct? It might make more sense for it to be:

```go
type GroupAddOptions struct {
    GID uint32
}
```

arichardet:
Feedback addressed.

PR #287, opened 2016-09-20
Add "Basic Usage" section to Server docs

### LINKED ISSUE #246: Add RPC auth documentation
When I started prototyping integration tests, I couldn't get the converge client to connect to a server that was running inside of a docker container. There is no documentation in converge for configuring this, and I'm not going to be the only one to run into that problem.

sehqlr:
The server documentation started off with special cases, and not with a
basic usage section at the top. I have added a first draft of this.

Fixes #246


docs_source/content/server.md
```diff
@@ -16,6 +16,32 @@ Converge comes with a server that can:
 - serve the Converge binary itself, for bootstrapping new systems inside your
   network
 
+## Basic Usage
+
+Install the converge binary to your server or device and run the `server` command.
```
BrianHicks: Let's have instructions here to get an actual token. The `uuid` utility produces a sufficiently random token, and is commonly available.

docs_source/content/server.md
```diff
@@ -16,6 +16,32 @@ Converge comes with a server that can:
...
+
+```shell
+converge server --rpc-token "YOLO security"
+```
+
+This will spin up a gRPC server on port 4774, with `YOLO security` set as the
+token string. You should see messages streaming from the server. If you run the
+server command without `--rpc-token`, then the output will include the
+generated token, and you can use that in place of `YOLO security` in future
+invocations.
```
BrianHicks: Could you mention that the token will be regenerated on every invocation of server if it's not set? It's worth emphasizing that this is really only for use with `--local`, although it does help get set up with a secured server quickly.

docs_source/content/server.md
```diff
@@ -16,6 +16,32 @@ Converge comes with a server that can:
...
+The next step is to run the converge binary in client mode. This can be on
+the same machine, or a different machine in your network. This example assumes
+that you have a HCL file called `your.hcl` that you wish to configure the
+server or device with.
+
+```shell
+converge plan --rpc-token "YOLO security" your.hcl
+```
+
+If your converge server is running on a different machine, then you will need
```
BrianHicks: … or on an address other than `localhost:4774`

docs_source/content/server.md
```diff
@@ -18,30 +18,33 @@ Converge comes with a server that can:
...
 ## Basic Usage
 
-Install the converge binary to your server or device and run the `server` command.
+To run converge in server mode with easy configuration, you just need
+the binary installed on your device, and a token to use for authenticating over
+RPC. In this example, we'll use the `uuid` utility to generate this token.
 
 ```shell
-converge server --rpc-token "YOLO security"
+token=`uuid`
```
BrianHicks: best to follow convention here of upper-casing the variable name

PR #288, opened 2016-09-21
Fix/239

rebeccaskinner:
Fix calls to `lookup` failing when the nodes are inside of a module.

### DIFF
--- a/load/dependencyresolver.go
+++ b/load/dependencyresolver.go
@@ -33,15 +33,16 @@ type dependencyGenerator func(node *parse.Node) ([]string, error)
 // the graph and creates edges to fit them
 func ResolveDependencies(ctx context.Context, g *graph.Graph) (*graph.Graph, error) {
 	logger := logging.GetLogger(ctx).WithField("function", "ResolveDependencies")
-	logger.Debug("resolving dependencies")
+	logger.Info("resolving dependencies")
 
-	return g.Transform(ctx, func(id string, out *graph.Graph) error {
+	r, e := g.Transform(ctx, func(id string, out *graph.Graph) error {
 		if id == "root" { // skip root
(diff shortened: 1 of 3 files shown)


load/dependencyresolver.go
```diff
@@ -33,9 +35,9 @@ type dependencyGenerator func(node *parse.Node) ([]string, error)
 // the graph and creates edges to fit them
 func ResolveDependencies(ctx context.Context, g *graph.Graph) (*graph.Graph, error) {
 	logger := logging.GetLogger(ctx).WithField("function", "ResolveDependencies")
-	logger.Debug("resolving dependencies")
+	logger.Info("resolving dependencies")
 
-	return g.Transform(ctx, func(id string, out *graph.Graph) error {
+	r, e := g.Transform(ctx, func(id string, out *graph.Graph) error {
```
ryane: I know it's been brought up before that single-letter variable names can make things hard to read. not sure it's a problem here since you are just returning them (don't even really need them).

load/dependencyresolver.go
```diff
@@ -33,9 +35,9 @@ type dependencyGenerator func(node *parse.Node) ([]string, error)
 // the graph and creates edges to fit them
 func ResolveDependencies(ctx context.Context, g *graph.Graph) (*graph.Graph, error) {
 	logger := logging.GetLogger(ctx).WithField("function", "ResolveDependencies")
-	logger.Debug("resolving dependencies")
+	logger.Info("resolving dependencies")
```
ryane: should this actually be an Info log instead of Debug?

BrianHicks: Yeah, looks like this changed back from the PR where I turned down the levels

Issue #199, opened 2016-08-30
RPC isn't used in healthchecks


BrianHicks:
RPC is used for `plan` and `apply`. It doesn't work for healthchecks, and apparently it needs a little bit of thought about how to do that.

rebeccaskinner:
Fixed in #289

PR #289, opened 2016-09-21
perform healthchecks over RPC

rebeccaskinner:
perform healthchecks over RPC


rebeccaskinner:
fixes #199

rpc/pb/root.proto
```diff
@@ -52,6 +52,14 @@ message DiffResponse {
 
 // Executor is responsible for remote execution on the machine
 service Executor {
+  // Healthcheck a module given by the location
+  rpc HealthCheck (LoadRequest) returns (stream StatusResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/machine/plan"
```
BrianHicks: conflicts with plan

rebeccaskinner: oops, copy-pasta error; fixed.

PR #290, opened 2016-09-21
docs: add draft resource authors guide

BrianHicks:


BrianHicks:
Code climate isn't going to like this but in this case all it's complaining about is that the H2s don't have an H1. We don't have H1s because Hugo inserts them automatically, so this isn't a problem.

BrianHicks:
Fixes #291

docs_source/content/resource-authors-guide.md
```diff
@@ -0,0 +1,131 @@
...
+`Preparer`).
+
+```go
+type Preparer struct {
+    Check string `hcl:"check"`
+    Apply string `hcl:"apply"`
+}
+
+func (p *Preparer) Render(r *resource.Renderer) (resource.Task, error) {
+    check, err := render.Render("check", p.Check)
```
rebeccaskinner: We should maybe mention what `render.Render` is doing here, and explain the expectation that render errors are returned.  It's maybe an API smell but right now we're reliant on the fact that a preparer will return `render.ErrUnresolvable` if that's what's returned from `render.Render` in order to understand when we need to defer node evaluation.

BrianHicks:
taken care of in 8544a29

PR #303, opened 2016-09-22
use thread-safe field cache

### LINKED ISSUE #302: preprocessor race condition
There is a race condition in the preprocessor because of the field cache

rebeccaskinner:
fixes #302


ryane:
lgtm on green. tests are passing consistently

PR #307, opened 2016-09-22
Fix/225 pipeline function refactor

rebeccaskinner:
refactor pipeline code to be idiomatic


apply/pipeline.go
```diff
@@ -50,114 +48,113 @@ func Pipeline(g *graph.Graph, id string, factory *render.Factory) executor.Pipel
...
-			}
-			result.PostCheck = plan.Status
-			if plan.HasChanges() {
-				result.Err = fmt.Errorf("%s still has changes after apply", g.ID)
-			}
-			return either.RightM(result)
-		})
+	val, pipelineError := plan.Pipeline(g.Graph, g.ID, g.RenderingPlant).Exec(task)
+	if pipelineError != nil {
+		return nil, val.(error)
```
BrianHicks: why is the value of `pipelineError` being discarded? Also, won't this cause a panic if `val` is not of the correct type?

rebeccaskinner: there probably should have been a test to catch that, looks like it's a vestige of the monadic code, it should be `return nil, pipelineError`

Issue #276, opened 2016-09-20
Allow ability to indicate group name when adding a user


arichardet:

sehqlr:
This was closed by #310, but not automatically. Closing manually.

PR #310, opened 2016-09-26
Add ability to indicate the group name when adding a user

arichardet:
Can now add a user by specifying either the group name or gid.

### DIFF
--- a/resource/user/user.go
+++ b/resource/user/user.go
@@ -34,22 +34,24 @@ const (
 
 // User manages user users
 type User struct {
-	UID      string
-	GID      string
-	Username string
-	Name     string
-	HomeDir  string
-	State    State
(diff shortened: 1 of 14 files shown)


arichardet:
#276

resource/user/user.go
```diff
@@ -93,10 +104,17 @@ func (u *User) Check(resource.Renderer) (resource.TaskStatus, error) {
 
 			switch {
 			case userByName != nil:
-				status.Level = resource.StatusFatal
```
BrianHicks: is this case no longer fatal?

arichardet: It should not have been fatal. With state=present, if the user exists, we do not change.

BrianHicks:
LGTM on green

PR #314, opened 2016-09-26
Add docker and vagrant integration test targets

sehqlr:
This prototype of an integration testing suite includes two platforms:
docker and vagrant. We are using docker for testing the same HCL file
against multiple linux OSes. We need vagrant because we need
an environment to test LVM modules, and it's a test of our plugin.

TODOs:

- More test cases than `basic.hcl`. What would be good candidates for
  this, besides all of the files?
- Can we get non-linux docker images to test, or is that going to need
  to be a separate thing?
- Right now, the docker target creates containers serially. I would like
  to add a target where we simulate a cluster with containers. Should we
  use docker compose for that? an HCL file? what?


docker-test.hcl
```diff
@@ -0,0 +1,83 @@
...
+  expose = ["4774"]
+
+  volumes = [
+    "{{param `converge-bin-dir`}}:/converge",
+    "{{env `PWD`}}/samples:/samples",
+    "{{env `PWD`}}/examples:/examples",
+  ]
+
+  entrypoint = ["/converge/converge"]
+  command    = ["server", "--rpc-token", "{{lookup `task.query.rpc-token.status.stdout`}}"]
```
BrianHicks: @rebeccaskinner is the fact that you have to lookup queries under the `status` path documented anywhere? Made me do a double take!

BrianHicks: To be clearer: I would have expected `{{lookup "task.query.rpc-token.stdout" }}` since it only has one field. `task.query.rpc-token.query.stdout` at most.

Vagrantfile
```diff
@@ -0,0 +1,14 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+	config.vm.box = "centos/7"
+	config.vm.synced_folder ".", "/vagrant", disabled: true
+	config.vm.synced_folder "./samples", "/converge_samples"
+	config.vm.provision :converge do |cvg|
```
ryane: if we're going to use vagrant for integration testing, we probably have to figure out a way to get the correct, local build of converge onto the vagrant machine. This might be best handled by a new option on the vagrant plugin.

also, this looks to be out of date with the latest version of the plugin.

docker-test.hcl
```diff
@@ -0,0 +1,83 @@
...
+}
+
+task.query "wait" {
+  query   = "sleep 10"
+  depends = ["docker.container.converge"]
+}
+
+task.query "container-port" {
+  query   = "docker inspect {{lookup `docker.container.converge.name`}} | jq '.NetworkSettings.Ports[\"4775/tcp\"][0].HostPort'"
```
sehqlr: Is there a way to escape the text/template input for when converge loads the command? It was giving me trouble before.

docker-test.hcl
```diff
@@ -0,0 +1,83 @@
...
+    "{{env `PWD`}}/samples:/samples",
+    "{{env `PWD`}}/examples:/examples",
+  ]
+
+  entrypoint = ["/converge/converge"]
+  command    = ["server", "--rpc-token", "{{lookup `task.query.rpc-token.status.stdout`}}"]
+}
+
+task.query "wait" {
+  query   = "sleep 10"
```
sehqlr: prolly not. How long does it take to run an image? 1 sec, 5?

docker-test.hcl
BrianHicks: ohhhh, that makes sense. No, we don't have a way to escape the double braces other than ugly stuff like `{{"{{"}}` :(

Vagrantfile
sehqlr: One of the things we need for the full-integration test is to check to make sure we have the latest version of the plugin. I haven't written that script yet

BrianHicks:
I’d make it as small a number as you can and incrementally adjust 
until it’s right. At least until @ryane’s wait.port lands.

quick-integration.hcl
```diff
@@ -0,0 +1,58 @@
+param "converge-bin-dir" {}
```
ryane: should we put this and the other testing scripts in a `testing` (or something) directory?

quick-integration.hcl
```diff
@@ -0,0 +1,58 @@
...
+
+param "image-name" {
+  default = "ubuntu"
+}
+
+param "image-tag" {
+  default = "latest"
+}
+
+docker.container "converge" {
```
ryane: you should use the docker.image resource to make sure that the image is present. this fails on a clean machine without that.

quick-integration.hcl
```diff
@@ -0,0 +1,58 @@
+param "converge-bin-dir" {}
+
+param "image-name" {
+  default = "ubuntu"
+}
+
+param "image-tag" {
+  default = "latest"
```
ryane: should use a specific version (`latest` is not a real version and while it is probably not likely with this particular workload, there is a risk that an updated ubuntu image could break the tests)

quick-integration.hcl
```diff
@@ -0,0 +1,58 @@
...
+  volumes = [
+    "/converge",
+    "{{env `PWD`}}/{{param `converge-bin-dir`}}:/converge/bin",
+    "{{env `PWD`}}/samples:/converge/samples",
+  ]
+
+  command = ["/converge/bin/converge", "server", "--no-token"]
+}
+
+task.query "wait" {
```
ryane: once #334 is merged, this could be replaced to wait for the rpc port to be available.

docker-test.hcl
BrianHicks: @ryane would the wait stuff solve this, do you think?

sehqlr:
I'm going to close this PR in favor of opening up a new one that doesn't have dup commits.

PR #317, opened 2016-09-26
Use text/tabwriter to align human output

### LINKED ISSUE #244: pretty printed changes should align values
the changes output arrows should line up. Currently they are staggered:

```
root/file.file1:
    Messages:
    Has Changes: yes
    Changes:
        content:
        before:
            
        after:
            this is a test
            of the content module
            
        gid: "" => "20"
        group: "" => "staff"
        type: "" => "file"
        uid: "" => "501"
        username: "" => "steve"
```

This is a minor issue, but reformat output so that `=>` arrows line up.

sehqlr:
This commit adds the text/tabwriter package to the human prettyprinter
package, and updates test case inputs. This aligns values based on a new
tab inserted to the right of the colon.

Fixes #244

### DIFF
--- a/prettyprinters/human/human.go
+++ b/prettyprinters/human/human.go
@@ -139,7 +139,7 @@ func (p *Printer) DrawNode(g *graph.Graph, id string) (pp.Renderable, error) {
 		return pp.HiddenString(), err
 	}
 
-	tabWriter := tabwriter.NewWriter(&out, 4, 4, 4, ' ', tabwriter.Debug)
+	tabWriter := tabwriter.NewWriter(&out, 1, 1, 1, ' ', 0)
 	_, err = tabWriter.Write(intermediate.Bytes())
 
 	return &out, err
@@ -173,7 +173,8 @@ func (p *Printer) diff(before, after string) (string, error) {
 	// remember when modifying these that diff is responsible for leading
 	// whitespace
 	if !strings.Contains(strings.TrimSpace(before), "\n") && !strings.Contains(strings.TrimSpace(after), "\n") {
-		return fmt.Sprintf("%q\t=>\t%q", strings.TrimSpace(before), strings.TrimSpace(after)), nil
+		sprintf := []string{"%q\t", "\x1b[1m", "=>", "\x1b[22m", "\t%q"}
+		return fmt.Sprintf(strings.Join(sprintf, ""), strings.TrimSpace(before), strings.TrimSpace(after)), nil
 	}
 
 	tmpl, err := p.template(`before:
--- a/prettyprinters/human/human.go
+++ b/prettyprinters/human/human.go
@@ -139,7 +139,7 @@ func (p *Printer) DrawNode(g *graph.Graph, id string) (pp.Renderable, error) {
 		return pp.HiddenString(), err
 	}
 
-	tabWriter := tabwriter.NewWriter(&out, 8, 8, 0, ' ', 0)
+	tabWriter := tabwriter.NewWriter(&out, 4, 4, 4, ' ', tabwriter.Debug)
 	_, err = tabWriter.Write(intermediate.Bytes())
 
 	return &out, err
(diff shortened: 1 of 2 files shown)


stevendborrelli:
output looks good:
```
root/file.directory.deeper:
        Messages:
        Has Changes: yes
        Changes:
                deeper:       "<absent>" => "<present>"
                deeper/a:     "<absent>" => "<present>"
                deeper/a/b:   "<absent>" => "<present>"
                deeper/a/b/c: "<absent>" => "<present>"
```

sehqlr:
This is going to need some manual reviews. I set the tabwriter.Debug flag, to make it easier to see what it's doing to the text. We need to decide the parameters to pass into the tabwriter struct, take a look at line 142 in the diff and the [tabwriter docs](https://golang.org/pkg/text/tabwriter/#Writer.Init).

BrianHicks:
@stevendborrelli that output actually looks completely fine to me. Are you saying you want it aligned on the arrows as well?

stevendborrelli:
@BrianHicks yes, it should be consistent. One test I had they were lined up, the other one they were not.

BrianHicks:
let me make sure I've got this. Do you want this?

```
root/docker.container.nginx:
        Messages:
        Has Changes: yes
        Changes:
                binds:             ""                          => ""
                command:           "nginx -g daemon off;"      => "nginx -g daemon off;"
                dns:               "8.8.8.8, 8.8.4.4"          => "8.8.8.8, 8.8.4.4"
                entrypoint:        ""                          => ""
                env:               "FOO=BAR"                   => "FOO=BAR"
                expose:            "443/tcp, 80/tcp, 8080/tcp" => "443/tcp, 80/tcp, 8080/tcp"
                image:             "nginx:1.10-alpine"         => "nginx:1.10-alpine"
                links:             ""                          => ""
                name:              "nginx-server"              => "nginx-server"
                ports:             "::80/tcp"                  => "::80/tcp"
                publish_all_ports: "false"                     => "false"
                status:            "exited"                    => "running"
                volumes:           ""                          => ""
                volumes_from:      ""                          => ""
                working_dir:       ""                          => ""

```

sehqlr:
Here's the output with the Debug |'s inserted:
screen shot 2016-09-29 at 11 21 40

BrianHicks:
hmm. Maybe we could make the arrow bright and then collapse that space? We're likely to run quite a bit over 80 characters if we align *everything*.

sehqlr:
We could also make the tabs narrower. Set the tabstop to 4 or 2 instead of 8

BrianHicks:
Doesn’t tabwriter convert to the smallest number of spaces anyway?

sehqlr:
There are controls for adding a number of spaces to each text cell before calculating anything else, as well as a minimum tab width.

BrianHicks:
What about a tab width of 1 and boldening the arrows?

sehqlr:
screen shot 2016-09-29 at 12 12 56

prettyprinters/human/human.go
```diff
@@ -173,7 +173,8 @@ func (p *Printer) diff(before, after string) (string, error) {
 	// remember when modifying these that diff is responsible for leading
 	// whitespace
 	if !strings.Contains(strings.TrimSpace(before), "\n") && !strings.Contains(strings.TrimSpace(after), "\n") {
-		return fmt.Sprintf("%q\t=>\t%q", strings.TrimSpace(before), strings.TrimSpace(after)), nil
+		sprintf := []string{"%q\t", "\x1b[1m", "=>", "\x1b[22m", "\t%q"}
```
BrianHicks: why are you doing the extra allocation for this?

BrianHicks: That'd be great. We don't need to allocate a list every time 👍

prettyprinters/human/human_test.go
```diff
@@ -197,7 +197,7 @@ func TestDrawNodeError(t *testing.T) {
 	testDrawNodes(
 		t,
 		Printable{"error": "x"},
-		"root:\n        Error: x\n        Messages:\n        Has Changes: yes\n        Changes:\n                error:   \"\" => \"x\"\n\n",
+		"root:\n Error: x\n Messages:\n Has Changes: yes\n Changes:\n  error: \"\" \x1b[1m=>\x1b[22m \"x\"\n\n",
```
BrianHicks: this should be using the same coloration logic as the rest of the human printer.

sehqlr:
You mean the terminal escape sequence?

BrianHicks:
Yep

sehqlr:
This approach bolds the whole diff, not just the arrow. The `bold` function should stay as is, because it's useful beyond this case, but if we _only_ want to bold the arrow, I have an idea for a solution.

BrianHicks:
yeah, it should only bold the error. Just add it as another field in the Sprintf call :)

sehqlr:
@BrianHicks @ryane This has been rebased and finished up. Could y'all double check my work and merge, plz?

Issue #110, opened 2016-07-22
proper arrays


BrianHicks:
Something like:

```hcl
module "package-install.hcl" "common-packages" {
  params {
    packages = [ "htop", "curl", "jq" ]
  }
}
```

Supported all the way through until rendering, when they would be turned into an appropriate value for a resource.

BrianHicks:
We're going to take it one further and inspect the values that the Preparers are expecting, then give it to them cleanly rather than relying on the Preparer implementors to correctly deserialize the fields and give them the right names.

PR #321, opened 2016-09-28
Refactor deserializers

BrianHicks:
This removes the need to render things in serializers. See the updated module author's guide for how the struct tags are now used.

Next step: letting users user more complex values in params.


BrianHicks:
Just to track: this is a partial fix for #110

PR #332, opened 2016-09-29
[WIP|NOMERGE] HCL metadata

sehqlr:
This is an experiment into implementing a metadata resource for converge files. It doesn't actually compile, and I don't know if we would even go this route. So, this is "fail early and fail often" philosophy at work.


BrianHicks:
We'll probably end up using this or something like it in Converge when we have a greater need for it (and oh is that day coming) but I'm going to close this PR for now. Thanks @sehqlr!

PR #331, opened 2016-09-29
Support for modifying linux groups

arichardet:
- Ability to modify a group name and/or gid

PR #333, opened 2016-09-29
Status Error Codes

### LINKED ISSUE #243: Execution Engine Ignoring Warning Levels
If a fatal error happens during a check or apply I do the following. 
```go
if err != nil {
			t.Status = &resource.Status{
				WarningLevel: resource.StatusFatal,
				Status:       err.Error(),
			}
			return t.Status, err
```
When I run `converge plan` expecting to see an error I get the following output
```
root/systemd.unit.test:
	Messages:
	Has Changes: no
	Changes: No changes
```

BrianHicks:
Improves error handling.

Fixes #243 
Required for #331 
~~Doesn't do documentation because it would conflict with #321 so that needs to maybe go in first?~~

### DIFF
--- a/apply/apply_test.go
+++ b/apply/apply_test.go
@@ -82,7 +82,7 @@ func TestApplyErrorsBelow(t *testing.T) {
 
 	errNode, ok := out.Get("root/err").(*apply.Result)
 	require.True(t, ok)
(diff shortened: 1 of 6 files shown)


stevendborrelli:
LGTM

PR #334, opened 2016-09-30
wait.query and wait.port resources

### LINKED ISSUE #238: ability to wait for a condition
In https://github.com/asteris-llc/converge-elk/blob/master/elk.hcl#L46, I am abusing `task.query` to wait for a service to become healthy. Should this be a native capability of converge?

It it could behave similarly to `task.query` (or possibly `healthcheck`) but with a built-in retry loop and timeout. Perhaps something like:

```hcl
task.probe "elasticsearch-health" {
   check = "[[ $(curl -s 'http://localhost:9200/_cluster/health' 2>/dev/null | jq -r .status) == \"green\" ]]"
   initialDelay = "1s"
   interval = "1s"
   timeout = "10s"
}
```

ryane:
fixes #238

### DIFF
--- a/docs_source/content/resources/wait.port.md
+++ b/docs_source/content/resources/wait.port.md
@@ -1,7 +1,7 @@
 ---
 title: "wait.port"
 slug: "wait-port"
-date: "2016-09-30T10:48:08-04:00"
+date: "2016-10-03T10:23:34-04:00"
 menu:
   main:
     parent: resources
@@ -16,7 +16,6 @@ menu:
 wait.port "8080" {
   host         = "localhost"
   port         = 8080
-  protocol     = "tcp"
   interval     = "1s"
   max_retry    = 10
   grace_period = "2s"
@@ -38,21 +37,24 @@ and the specified Port.
 
 - `interval` (duration string)
 
-  the amount of time to wait in between checks. The format is Go's duraction
+  the amount of time to wait in between checks. The format is Go's duration
 string. A duration string is a possibly signed sequence of decimal numbers,
 each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or
-"2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+"2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". If
+the interval is not specified, it will default to 5 seconds.
 
 - `grace_period` (duration string)
 
   the amount of time to wait before running the first check and after a
-successful check. The format is Go's duraction string. A duration string is
+successful check. The format is Go's duration string. A duration string is
 a possibly signed sequence of decimal numbers, each with optional fraction
 and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units
-are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+are "ns", "us" (or "µs"), "ms", "s", "m", "h". If no grace period is
+specified, no grace period will be taken into account.
 
 - `max_retry` (int)
 
-  the maximum number of attempts before the wait fails.
+  the maximum number of attempts before the wait fails. If the maximum number
+of retries is not set, it will default to 5.
 
 
--- a/docs_source/content/resources/wait.port.md
+++ b/docs_source/content/resources/wait.port.md
@@ -1,0 +1,58 @@
+---
+title: "wait.port"
+slug: "wait-port"
+date: "2016-09-30T10:48:08-04:00"
+menu:
+  main:
+    parent: resources
+---
+
+
+
+
+## Example
+
+```hcl
+wait.port "8080" {
+  host         = "localhost"
+  port         = 8080
+  protocol     = "tcp"
+  interval     = "1s"
+  max_retry    = 10
+  grace_period = "2s"
+}
+
+```
+
+
+## Parameters
+
+- `host` (string)
+
+  a host name or ip address. A TCP connection will be attempted at this host
+and the specified Port.
+
+- `port` (required int)
+
+  the TCP port to attempt to connect to.
+
+- `interval` (duration string)
+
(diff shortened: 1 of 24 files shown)


docs_source/content/resources/wait.port.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+
+## Example
+
+```hcl
+wait.port "8080" {
+  host         = "localhost"
+  port         = 8080
+  protocol     = "tcp"
```
BrianHicks: this isn't in the docs, which means it's not in your code, right?

examples/elk/Vagrantfile
```diff
@@ -1,7 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-converge_version = "0.2.0"
+converge_version = "0.3.0"
```
BrianHicks: what's this for? We haven't released 0.3.0 yet.

examples/elk/converge/elk.hcl
```diff
@@ -50,26 +51,22 @@ task "filebeat-enable" {
...
-MAX_SECONDS=60
-while /bin/true
-do
-    status=$(curl -s 'http://localhost:9200/_cluster/health' 2>/dev/null | jq -r .status)
-    if [ "$status" == "yellow" ] || [ "$status" == "green" ] ; then
-        exit 0
-    fi
-    [[ "$SECONDS" -ge "$MAX_SECONDS" ]] && exit 1
-done
+wait.query "elasticsearch-wait" {
```
BrianHicks: :clap: :clap: :clap: :clap: :clap: :clap:

resource/wait/port/port.go
```diff
@@ -0,0 +1,90 @@
...
+	Port int
+}
+
+// Check if the port is open
+func (p *Port) Check(resource.Renderer) (resource.TaskStatus, error) {
+	p.Status = resource.NewStatus()
+
+	alive, err := p.checkConnection()
+	if err != nil {
+		return p, errors.Wrapf(err, "failed to check connection")
```
BrianHicks: shouldn't this be more like "connection was not open by the given timeout"? The code *did* check the connection, it just potentially timed out.

resource/wait/port/preparer.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// Preparer handles wait.query tasks
+type Preparer struct {
+	// a host name or ip address. A TCP connection will be attempted at this host
+	// and the specified Port.
+	Host string `hcl:"host"`
+
+	// the TCP port to attempt to connect to.
+	Port int `hcl:"port" required:"true"`
```
BrianHicks: doesn't this need to be non-zero? Should that check be added?

resource/wait/port/preparer.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// the amount of time to wait in between checks. The format is Go's duraction
+	// string. A duration string is a possibly signed sequence of decimal numbers,
+	// each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or
+	// "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	Interval string `hcl:"interval" doc_type:"duration string"`
```
BrianHicks: what are the default values for these? What happens if this is an empty string?

resource/wait/port/preparer.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// the amount of time to wait before running the first check and after a
+	// successful check. The format is Go's duraction string. A duration string is
+	// a possibly signed sequence of decimal numbers, each with optional fraction
+	// and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units
+	// are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	GracePeriod string `hcl:"grace_period" doc_type:"duration string"`
+
+	// the maximum number of attempts before the wait fails.
+	MaxRetry int `hcl:"max_retry"`
```
BrianHicks: I don't see a default for this either. Doesn't that make it zero?

docs_source/content/resources/wait.port.md
ryane: whoops, I'll remove it. I added that before I realized that udp checks are not really feasible.

examples/elk/Vagrantfile
ryane: I figured it was better to set that now since the examples are built to download the release binaries and will no longer work with the 0.2.0 binary. All in all, I think we need a better system for the examples. I will open an issue on that.

resource/wait/port/port.go
ryane: actually this means something other than a network error (net.OpErr) occurred. I don't have an example of what this may be at this point as I've never seen anything else yet and don't see any other error types being returned after a cursory look through the stdlib code. If the connection fails b/c of an OpErr, checkConnection returns false and nil for the error.

ryane:
issues addressed and rebased. ready for review again

resource/wait/port/port.go
```diff
@@ -0,0 +1,96 @@
...
+func (p *Port) Check(resource.Renderer) (resource.TaskStatus, error) {
+	p.Status = resource.NewStatus()
+
+	err := p.CheckConnection()
+	if err == nil {
+		if p.RetryCount > 0 {
+			p.Status.AddMessage(fmt.Sprintf("Passed after %d retries (%v)", p.RetryCount, p.Duration))
+		}
+	} else {
+		p.RaiseLevel(resource.StatusWillChange)
```
ryane: The desired state is that the port is listening. So when you run plan, StatusWillChange tells you that, optimistically, the port will be listening after the apply. Also, it gives you the color in the converge output that lets the user more easily see whether or not the port is available during the plan phase.

That said, I don't think leaving the status as StatusNoChange will affect the functionality outside of the converge output. Do you think that is preferred?

PR #336, opened 2016-09-30
plan: test, which illustrate "return nil, err" problem

avnik:


helpers/faketask/faketask.go
```diff
@@ -22,46 +22,64 @@ import (
 
 // FakeTask for testing things that require real tasks
 type FakeTask struct {
-	Status string
-	Level  resource.StatusLevel
-	Error  error
+	Status      string
+	Level       resource.StatusLevel
+	Error       error
+	NilAndError bool
```
rebeccaskinner: I wonder if it would make more sense to have a `type NilTask struct` that just returns `nil, error` to keep the code clean

avnik:
@rebeccaskinner I add commit with fix, ad can rework test, if you or @BrianHicks  prefer it.

This fix need to be merged before lvm branch

plan/pipeline.go
```diff
@@ -108,6 +108,11 @@ func (g *pipelineGen) PlanNode(taski interface{}) (interface{}, error) {
 	}
 	status, err := twrapper.Task.Check(renderer)
 
+	// create empty Status structure, if it not created in .Check()
```
ryane: if we use this pr, we probably need the same check in apply/pipeline.go. That is where I am panicking in #343

avnik: @ryane Will do, so don't panic ;)

helpers/faketask/faketask.go
```diff
@@ -65,6 +65,25 @@ func WillChange() *FakeTask {
...
+func (*NilTask) Check(resource.Renderer) (resource.TaskStatus, error) {
+	return nil, errors.New("check error")
+}
+
+// Always raise error
+func (*NilTask) Apply() (resource.TaskStatus, error) {
+	return nil, errors.New("apply error")
+}
+
+// NilAndError return a FakeTask that will simulate `return nil, err` cawe
```
ryane: case

apply/apply_test.go
```diff
@@ -105,6 +105,20 @@ func TestApplyStillChange(t *testing.T) {
...
 
+func TestApplyNilError(t *testing.T) {
+	defer logging.HideLogs(t)()
+
+	g := graph.New()
+	g.Add("root", &plan.Result{Status: &resource.Status{Level: resource.StatusWillChange}, Task: faketask.NilAndError()})
+
+	require.NoError(t, g.Validate())
+
+	// applying should return error, and not panicking
```
ryane: // Apply should return an error and not panic

apply/apply_test.go
```diff
@@ -105,6 +105,20 @@ func TestApplyStillChange(t *testing.T) {
 	}
 }
 
+func TestApplyNilError(t *testing.T) {
```
ryane: add comments on exported methods (will fix cc issues)

plan/plan_test.go
```diff
@@ -45,6 +45,21 @@ func TestPlanNoOp(t *testing.T) {
 	assert.Equal(t, task, result.Task)
 }
 
+func TestPlanNilAndError(t *testing.T) {
```
ryane: add comments on exported methods

helpers/faketask/faketask.go
```diff
@@ -65,6 +65,25 @@ func WillChange() *FakeTask {
 	}
 }
 
+// Task which always return (nil, error) tuple
```
ryane: fix comments on these. see https://codeclimate.com/github/asteris-llc/converge/pull/336

avnik:
@rebeccaskinner @ryane  all test are green now.
If Brian offer better fix, we can revert commit with my fixups, and leave tests only.

PR #339, opened 2016-09-30
Ability to use pointers as a preparer value

### LINKED ISSUE #337: validate non-empty string values on some resources
After #321, it looks like we have to add back some validation in Preparers to ensure that some string fields are not empty. 
- docker.container: name, image
- group: name
- user: username

These fields are required but they also should not be empty. I have confirmed that the docker.container fields are definitely a problem and it _looks_ like user and group will be a problem as well. For example, converge does not complain about this:

``` hcl
docker.container "nginx" {
  name  = "nginx"
  image = ""
}
```

but trying to apply this will obviously fail so we should add further validations.

### LINKED ISSUE #338: handle parameters with valid zero value
After #321, we need to make a change to handle parameters for which zero is a valid value. Zero is a valid value for user.user gid and uid parameters, and user.group gid parameter. A user can be created by providing a uid or allowing it to be set for you in the system call. 

The problem now is that if the parameter is not provided in the configuration, the preparer sets its value to zero.

BrianHicks:
Fixes #338

### DIFF
--- a/docs_source/content/resource-authors-guide.md
+++ b/docs_source/content/resource-authors-guide.md
@@ -110,6 +110,42 @@ To get values other than strings (bools, ints, et cetera), you just need to
 specify them. Converge will render the values and parse them from strings, if
 necessary.
 
+### Zero Values
+
+Sometimes you need to disambiguate between a zero value the user provided and
+one that Go did. In this case, use a pointer to that type. For example, your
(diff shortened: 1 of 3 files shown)


arichardet:
LGTM!

stevendborrelli:
Also works for me.

PR #340, opened 2016-09-30
Add Compound Parameters

### LINKED ISSUE #110: proper arrays

BrianHicks:
See examples for more.

Fixes #110

### DIFF
--- a/docs_source/content/params-and-templates.md
+++ b/docs_source/content/params-and-templates.md
@@ -75,7 +75,7 @@ resource to the named parameter.
 {{< note title="Backwards arguments?" >}}
 Some of the utility functions (like `split` and `join`) seem to have their
 arguments backwards. This is to allow you to pipe values to them. For example,
-joining a list on strings is `{{paramList "x" | join "\n"}}`
+joining a list of strings is `{{paramList "x" | join "\n"}}`
 {{< /note >}}
 
 - **split** splits a string (second argument) at the instances of another string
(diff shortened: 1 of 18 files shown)


BrianHicks:
@stevendborrelli that TODO codeclimate is complaining about is the one that it'd be really helpful if you'd fill in

docs_source/content/params-and-templates.md
```diff
@@ -0,0 +1,87 @@
...
+
+- **env** retrieves an item (named by the first argument) from an environment
+  variable
+
+### Utility
+
+{{< note title="Backwards arguments?" >}}
+Some of the utility functions (like `split` and `join`) seem to have their
+arguments backwards. This is to allow you to pipe values to them. For example,
+joining a list on strings is `{{paramList "x" | join "\n"}}`
```
ryane: "list of strings"?

BrianHicks: Oops! I've fixed this in the latest push.

docs_source/content/params-and-templates.md
```diff
@@ -0,0 +1,87 @@
...
+  variable
+
+### Utility
+
+{{< note title="Backwards arguments?" >}}
+Some of the utility functions (like `split` and `join`) seem to have their
+arguments backwards. This is to allow you to pipe values to them. For example,
+joining a list of strings is `{{paramList "x" | join "\n"}}`
+{{< /note >}}
+
```
BrianHicks: The answer before: join fails because it's the wrong type. But that's just silly, since we don't have any way to do type conversions. So now it'll just stringify any arguments passed to it. See https://github.com/asteris-llc/converge/pull/340/commits/7e65cc64984a596f0da6fab6ee304c5cdce89746

PR #349, opened 2016-10-04
Feature: rpm/apt module

Zogg:
This is my first attempt at writing a new module. I haven't read any docs, looked at other modules to understand/deduce logic and this is the outcome. I expect pointers/directions and comments on pretty much ever block of the code - dont hesitate. ;)

My first impressions are such:
1) I need to know a LOT about converge internals to write a module
2) hcl syntax is too redundant, reminds of CF-Engine's 
3) the speed and rpc architecture is nice


resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/asteris-llc/converge/resource"
+)
+
+// Package is an API for package state
+type Package struct {
+	resource.TaskStatus
+
+	Name  string
+	State string
```
arichardet: I suggest making State a type: `type State string` then having constants for the different states. You can look at user.user or user.group if you want to see this done elsewhere.

resource/packages/apt/preparer.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Preparer for Content
+//
+// Content renders content to disk
+type Preparer struct {
+	// Content is the file content. This will be rendered as a template.
+	Name string `hcl:"name"`
+
+	// Destination is the location on disk where the content will be rendered.
+	State string `hcl:"state"`
```
arichardet: If either of these parameters are required, you should add the `required:"true"` tag. You should also include a `valid_values:"present,absent"` tag for State

resource/packages/apt/preparer.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/asteris-llc/converge/load/registry"
+	"github.com/asteris-llc/converge/resource"
+)
+
+// Preparer for Content
+//
+// Content renders content to disk
+type Preparer struct {
+	// Content is the file content. This will be rendered as a template.
+	Name string `hcl:"name"`
```
arichardet: If this parameter is required, you should add the `required:"true"` tag.

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Check if the package has to be 'installed', or 'absent'
+func (p *Package) Check(resource.Renderer) (resource.TaskStatus, error) {
+	status := resource.NewStatus()
+
+	currentPkgStatusRaw, err := exec.Command("bash", "-c", "apt-cache policy "+p.Name+" | grep Installed | awk '{print $2}'").Output()
```
BrianHicks: Can we use `sh` here instead? Bash is not guaranteed to be installed (e.g. in a container)

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		log.WithField("module", "apt").Warn(err)
```
BrianHicks: this should return the error instead of logging it and continuing. Maybe this:

```go
return status, errors.Wrapf(err, "checking package %s", p.Name)
```

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	currentPkgStatus := strings.TrimSpace(string(currentPkgStatusRaw))
+
+	log.WithField("module", "apt").WithField("exec", "bash").Info(string(currentPkgStatus))
```
BrianHicks: this should be a `Debug` level, if included at all. We don't log much in these functions.

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+	log.Info(p)
```
BrianHicks: this one should just go

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	p.TaskStatus = status
+	return p, nil
+}
+
+// Apply desired package state
+func (p *Package) Apply() (resource.TaskStatus, error) {
+	if p.State == "installed" {
+		currentPkgStatus, err := exec.Command("bash", "-c", "apt-get install -y "+p.Name).Output()
```
BrianHicks: `sh` here too. Also, can we use the `--long` flags? It's easier to maintain in the future. I think this one is either `--yes` or `--assume-yes`.

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			log.WithField("module", "apt").Warn(err)
```
BrianHicks: more error returns here, please

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+			log.WithField("module", "apt").Warn(string(currentPkgStatus))
+		}
+
+		log.WithField("module", "apt").Info(fmt.Sprintf("Installed %q: \n %q", p.Name, string(currentPkgStatus)))
```
BrianHicks: this should be a difference, not log output.

I'm getting uppity about using logs to display information to the user because the server running this code and the client requesting it will probably be attached to different terminals, at minimum. Probably even separate machines. If you need to add status, check out the messages field on the status struct.

resource/packages/apt/apt.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			log.WithField("module", "apt").Warn(err)
+			log.WithField("module", "apt").Warn(string(currentPkgStatus))
+		}
+
+		log.WithField("module", "apt").Info(fmt.Sprintf("Removed %q: \n %q", p.Name, string(currentPkgStatus)))
+	}
+
+	status := resource.NewStatus()
+	status.RaiseLevel(resource.StatusNoChange)
```
BrianHicks: as written, this will return a `StatusNoChange` even if something changes.

resource/packages/apt/preparer.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/asteris-llc/converge/load/registry"
+	"github.com/asteris-llc/converge/resource"
+)
+
+// Preparer for Content
+//
+// Content renders content to disk
+type Preparer struct {
```
BrianHicks: how does this whole thing handle versioning?

resource/packages/apt/preparer.go
BrianHicks: This could be a string alias type too, that would make validation easier.

resource/packages/apt/apt.go
Zogg: I'd follow the folklore when it comes to flags. i.e. I'v never seen, even in stackoverflow folkland, `apt-get install -y` with long flag, even man page shows `-y`. :)

resource/packages/apt/apt.go
Zogg: But there is no `StatusHasChanged`, the closest one is `StatusWillChange`. But returning that from Apply() doesnt make sense. It has already changed, hence no change is expected anymore.

resource/packages/apt/apt.go
```diff
@@ -35,14 +34,12 @@ type Package struct {
...
-		log.WithField("module", "apt").Warn(err)
+		return status, errors.Wrapf(err, "checking package %s", p.Name)
 	}
 	currentPkgStatus := strings.TrimSpace(string(currentPkgStatusRaw))
 
-	log.WithField("module", "apt").WithField("exec", "bash").Info(string(currentPkgStatus))
-	log.Info(p)
 	if p.State == "installed" {
 		if string(currentPkgStatus) == "(none)" {
 			status.AddDifference(p.Name, string(currentPkgStatus), "installed", "")
```
arichardet: You should set `status.Level = resource.StatusWillChange` in Check() anywhere there will be changes, i.e. anywhere you call `status.AddDifference()`.

resource/rpm/rpm.go
```diff
@@ -0,0 +1,88 @@
...
+
+	currentPkgStatusRaw, err := exec.Command("sh", "-c", fmt.Sprintf("yum info %q | grep Repo | awk '{print $3}'", p.Name)).Output()
+	if err != nil {
+		return status, errors.Wrapf(err, "checking package %s", p.Name)
+	}
+	currentPkgStatus := strings.TrimSpace(string(currentPkgStatusRaw))
+
+	if p.State == "installed" {
+		if string(currentPkgStatus) != "installed" {
+			status.AddDifference(p.Name, string(currentPkgStatus), "installed", "")
```
arichardet: You should set `status.Level = resource.StatusWillChange` in Check() anywhere there will be changes, i.e. anywhere you call `status.AddDifference()`.

resource/packages/apt/preparer.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Preparer for APT Package
+//
+// APT Package manages system packages
+type Preparer struct {
+	// Name of the system package to be managed.
+	Name string `hcl:"name" required:"true" `
+
+	// State defines desired system package state.
+	State string `hcl:"state" valid_values:"installed,absent" default:"installed"`
```
arichardet: Should probably use the most common values for state, or allow for all: present, installed, latest, absent, removed; or at least use the terms that go together like present and absent or installed and removed (instead of mixing installed and absent).

resource/packages/apt/apt.go
Zogg: I haven't dug deeper: does `AddDifference()` set some kind of Status on its own?

stevendborrelli:
Please improve error handling. For example, if I run as non root the install fails, but doesn't expose this information to the user. 

```
2016-10-07T15:12:32Z |INFO| Installed "nc": 
 "Loaded plugins: fastestmirror\n"	module=rpm
2016-10-07T15:12:33Z |INFO| 	exec=bash module=rpm
2016-10-07T15:12:33Z |INFO| &{0xc4201c0f00 nc installed}	
2016-10-07T15:12:34Z |INFO| installed	exec=bash module=rpm
2016-10-07T15:12:34Z |INFO| &{<nil> gcc installed}	
2016-10-07T15:12:34Z |INFO| got status	component=client file=samples/rpm.hcl id=root run=STARTED stage=APPLY

root/rpm.package.mc:
	Error: root/rpm.package.mc still has changes after apply
	Messages:
	Has Changes: yes
	Changes:
		nc: "" => "installed"

 * root/rpm.package.mc: root/rpm.package.mc still has changes after apply
 * root: error in dependency "root/rpm.package.mc"
```

BrianHicks:
Thanks for this @zogg, we're going to close it for now. We'll probably end up using most of it in a different form. Thanks again! ⛵️

Issue #348, opened 2016-10-04
Tidy up codebase by fixing `make lint` reports


sehqlr:
As of posting this issue, `make lint` returns this report:
```
# golint
resource/group/group_default.go:26:1: comment on exported method System.AddGroup should be of the form "AddGroup ..."
resource/group/group_default.go:32:1: comment on exported method System.DelGroup should be of the form "DelGroup ..."
resource/group/group_default.go:38:1: comment on exported method System.LookupGroup should be of the form "LookupGroup ..."
resource/group/group_default.go:44:1: comment on exported method System.LookupGroupID should be of the form "LookupGroupID ..."
resource/group/group_default.go:26:1: comment on exported method System.AddGroup should be of the form "AddGroup ..."
resource/group/group_default.go:32:1: comment on exported method System.DelGroup should be of the form "DelGroup ..."
resource/group/group_default.go:38:1: comment on exported method System.LookupGroup should be of the form "LookupGroup ..."
resource/group/group_default.go:44:1: comment on exported method System.LookupGroupID should be of the form "LookupGroupID ..."
./rpc/jwt.go:119:1: exported method JWTAuth.Protect should have comment or be unexported
./rpc/pb/mocks/Grapher_GraphServer.go:11:6: don't use underscores in Go names; type Grapher_GraphServer should be GrapherGraphServer
# go tool vet
cmd/apply.go:60: declaration of "err" shadows declaration at cmd/apply.go:55
cmd/apply.go:139: declaration of "err" shadows declaration at cmd/apply.go:88
cmd/graph.go:67: declaration of "err" shadows declaration at cmd/graph.go:62
cmd/healthcheck.go:61: declaration of "err" shadows declaration at cmd/healthcheck.go:56
cmd/key.go:80: declaration of "err" shadows declaration at cmd/key.go:64
cmd/plan.go:60: declaration of "err" shadows declaration at cmd/plan.go:55
cmd/plan.go:139: declaration of "err" shadows declaration at cmd/plan.go:88
keystore/keystore_test.go:39: declaration of "err" shadows declaration at keystore/keystore_test.go:26
keystore/keystore_test.go:42: declaration of "err" shadows declaration at keystore/keystore_test.go:26
load/nodes.go:74: declaration of "err" shadows declaration at load/nodes.go:59
make: *** [lint] Error 1
```
This is a small collection of easy changes, and perfect for a new contributor.

sehqlr:
[...]

PR #350, opened 2016-10-05
Fix golint and govet warnings

Zacho2:
These changes Address #348. It seems that `go vet` with the shadow option is causing a false positive for this warning: `load/nodes.go:74: declaration of "err" shadows declaration at load/nodes.go:59` so it remains.


BrianHicks:
Boo, looks like I merged it in a way that GitHub didn't pick up. We've got your changes in at https://github.com/asteris-llc/converge/commit/c29ac9b8fbeaa3af25c13a98e03fa35349a4af47

Thanks again! 👍

BrianHicks:
If you're participating in Hacktoberfest, this still counts. (if I'm reading the rules correctly)

PR #352, opened 2016-10-05
Feature/param name in error

### LINKED ISSUE #335: "param is required" error should include param name
If a param is required and does not have a default value, the converge command
  will fail with a `param is required` error. While the error message does include
  the specific hcl file, it does not tell you which param is missing. It would be
  better if the error included the param name. This would be helpful on files that may
  have several params or in the scenario where the user just made a small typo in
  the param name on the command line (speaking from experience).

  Example:

  ```shell
$ ./converge graph --local examples/docker-swarm-mode/converge/worker.hcl
2016-09-30T12:43:27-04:00 |INFO| serving        addr=:47740 component=rpc
2016-09-30T12:43:27-04:00 |ERROR| could not render      component=rpc error=param is required location=examples/docker-swarm-mode/converge/worker.hcl runID=7b76007f-0aab-40c6-b4ab-aa1087e16212
2016-09-30T12:43:27-04:00 |ERROR| loading failed        component=rpc error=rendering examples/docker-swarm-mode/converge/worker.hcl: param is required function=grapher.Graph runID=7b76007f-0aab-40c6-b4ab-aa1087e16212
2016-09-30T12:43:27-04:00 |FATAL| could not get graph   component=client error=error getting graph component: rpc error: code = 2 desc = loading failed: rendering examples/docker-swarm-mode/converge/worker.hcl: param is required file=examples/docker-swarm-mode/converge/worker.hcl
  ```

ryane:
fixes #335

### DIFF
--- a/fetch/resolve.go
+++ b/fetch/resolve.go
@@ -23,7 +23,7 @@ import (
 
 // ResolveInContext resolves a path relative to another
 func ResolveInContext(loc, ctx string) (string, error) {
-	log.WithField("target", loc).WithField("context", ctx).Debug("resolving target in contextj")
+	log.WithField("target", loc).WithField("context", ctx).Debug("resolving target in context")
 
 	var (
 		locScheme, locPath = parse(loc)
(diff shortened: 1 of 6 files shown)


ryane:
this might be a naive implementation but does seem to solve the problem. let me know if you think there is a better way to tackle this

PR #355, opened 2016-10-06
[Updated] Add quick and full integration test targets

sehqlr:
This commit adds integration test code in two phases:
- quick-integration.hcl, for CI builds. Run with `make integration`
- full-integration.sh, for nightly builds. This is triggered by setting
  CVG_FULL_INTEGRATION in the environment

We are using docker for testing the same HCL file
against multiple linux OSes. We need vagrant because we need
an environment to test LVM modules, and it's a test of our plugin.

Future Work:
- Port blackbox tests to integration tests where appropriate
- Set up nightly build machine, and add integration test to wercker
- Right now, the docker target creates containers serially. I would like
  to add a target where we simulate a cluster with containers. Should we
  use docker compose for that? an HCL file? what?


testing/full-integration.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+cd $(dirname "$0")
```
ryane: untested but maybe try something like `cd "$(dirname "$0")" || exit 1` to cleanup code climate issues

Makefile
```diff
@@ -129,4 +130,13 @@ docs: docs_source/**/*
 	$(MAKE) -C docs_source
 	mv docs_source/public docs
 
-.PHONY: test gotest vendor-update vendor-clean xcompile package samples/errors/*.hcl blackbox/*.sh lint bench license-check
+integration: xcompile
+	@echo
+	@echo === quick integration test ===
+	$(INTEGRATION_CMD) testing/quick-integration.hcl
+	@ if [ -n "$$CVG_FULL_INTEGRATION" ];\
```
ryane: I think we should probably document the use of this env var somewhere. maybe in CONTRIBUTING.md?

Vagrantfile
```diff
@@ -0,0 +1,14 @@
...
+	config.vm.box = "centos/7"
+	config.vm.synced_folder ".", "/vagrant", disabled: true
+	config.vm.synced_folder "./samples", "/converge_samples"
+	config.vm.provision :converge do |cvg|
+		cvg.bikeshed = [
+			"/converge_samples/basic.hcl"
+		]
+		cvg.install = true
+	end
+end
```
ryane: this is out of date with the most recent version of the Vagrant plugin.

```
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vagrant-converge:
* The following settings shouldn't exist: bikeshed
* No value provided for hcl


There are errors in the configuration of this machine. Please fix
the following errors and try again:

vagrant-converge:
* The following settings shouldn't exist: bikeshed
* No value provided for hcl
```

Also, effective testing using vagrant is blocked by #356. I wonder if we should just leave this part out of the PR until we solve those problems.

Makefile
sehqlr: I'll add a note there, good catch

BrianHicks:
Thanks for this, Sam. We'll probably use this as a basis for integration testing in the future but I'm going to close it for now so we have a clearer view of our open PRs.

PR #363, opened 2016-10-07
cmd/server.go: use errgroup instead of waitgroup

### LINKED ISSUE #295: Use errgroup in server
Probably an improvement over what we're doing now and it has context support. https://godoc.org/golang.org/x/sync/errgroup#ex-Group--Pipeline

QuentinPerez:
closes #295

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -16,15 +16,15 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"os"
-	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/asteris-llc/converge/rpc"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/grpclog"
 )
 
@@ -54,8 +54,7 @@ var serverCmd = &cobra.Command{
 		ctx, cancel := context.WithCancel(context.Background())
 		GracefulExit(cancel)
 
-		var running sync.WaitGroup
-		running.Add(2)
+		group, ctx := errgroup.WithContext(ctx)
 
 		setLocal(true) // so we generate a token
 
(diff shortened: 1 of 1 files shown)


cmd/server.go
```diff
@@ -77,39 +78,34 @@ var serverCmd = &cobra.Command{
...
 				ctx,
 				getRPCAddr(),
 				sslConfig,
 				viper.GetString("root"),
 				viper.GetBool("self-serve"),
-			)
-			if err != nil {
-				log.WithError(err).Fatal("could not run RPC")
+			); err != nil {
+				return fmt.Errorf("%v: could not run RPC", err)
```
BrianHicks: We use `errors.Wrap` from `github.com/pkg/errors` instead of `fmt.Errorf` to avoid losing information. Could you please change it? The code here should look like `errors.Wrap(err, "could not run RPC")`. Same (but with different error message) below.

cmd/server.go
```diff
@@ -127,13 +123,15 @@ var serverCmd = &cobra.Command{
...
 			}
-
 			httpLog.Info("halted")
-		}()
+			return
+		})
 
-		running.Wait()
+		if err = group.Wait(); err != nil {
+			log.Fatal(err)
```
BrianHicks: Could you change this to `log.WithError(err).Fatal("appropriate message")`? Maybe something like "serving failed"?

cmd/server.go
```diff
@@ -54,8 +56,7 @@ var serverCmd = &cobra.Command{
 		ctx, cancel := context.WithCancel(context.Background())
 		GracefulExit(cancel)
 
-		var running sync.WaitGroup
-		running.Add(2)
+		var group errgroup.Group
```
BrianHicks: Let's change this to use [`errgroup.WithContext`](https://godoc.org/golang.org/x/sync/errgroup#WithContext) so we can do proper cleanup. Assigning it over the existing `ctx` variable is fine.

BrianHicks:
Oh and if you have any questions you want answered right away, feel free to message me in the [Converge Slack](http://converge-slack.aster.is)

QuentinPerez:
done

BrianHicks:
Yep, looks great. Thanks again! :tada:

PR #364, opened 2016-10-11
Feature/converge bootstrap

stevendborrelli:
Add modules that set up build environments on various operating systems.


examples/converge-build/readme.md
```diff
@@ -0,0 +1,80 @@
+# converge-builder
```
BrianHicks: nitpick: should have the same title as the name of the directory

PR #366, opened 2016-10-11
Add testing section to contribution guidelines

arichardet:
Add information on testing and including it when contributing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Hi! Thanks for being interested in contributing to Converge. We're really happy
 to have you (we have mini-parties in Slack whenever someone external opens a
-PR!)
+PR!).
 
 Here's a quick guide for how you should expect the contribution process to go.
 Above all, the core team and everyone involved with Converge is expected to
@@ -15,6 +15,9 @@ follow the [code of conduct](CODE_OF_CONDUCT.md).
     - [Filing Bugs and Getting Help](#filing-bugs-and-getting-help)
     - [Feature Requests](#feature-requests)
         - [Contributing Features and Fixes](#contributing-features-and-fixes)
+            - [Test Coverage](#test-coverage)
+                - [Our Testing Goals](#our-testing-goals)
+                - [Tests](#tests)
             - [Documentation](#documentation)
     - [Code of Conduct](#code-of-conduct)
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 Hi! Thanks for being interested in contributing to Converge. We're really happy
 to have you (we have mini-parties in Slack whenever someone external opens a
-PR!)
+PR!).
```
BrianHicks: we don't need the additional trailing periods here and in the contribution requirements

CONTRIBUTING.md
```diff
@@ -55,13 +58,34 @@ often much easier to reason about than describing a problem and solution.
...
   first so we can discuss.
 
+#### Test Coverage
+
+##### Our Testing Goals
+
+We want to be sure of a few things, and testing will help us accomplish this:
+
+- Help ensure a new feature or changes to an existing feature are properly
+implemented.
```
BrianHicks: adding two spaces before this line will fix code climate complaining

CONTRIBUTING.md
```diff
@@ -55,13 +58,34 @@ often much easier to reason about than describing a problem and solution.
...
+
+##### Tests
+
+Adequate testing can vary based on the change being introduced. If you have
+questions, don't hesitate to ask. In general, this is how we'd like tests to be
+addressed:
+
+- Include tests for the way your code interacts with the core engine.
+- Test functionality of the code you have introduced, say in a new module.
+- Mock interaction with the system.
```
BrianHicks: This is not always true. In fact, I'd say we should avoid mocking when possible. Too easy to get out of sync between mocks and reality.

BrianHicks: This list should also include something like "tests that demonstrate a bug, and that the bug has been resolved."

CONTRIBUTING.md
```diff
@@ -55,13 +58,34 @@ often much easier to reason about than describing a problem and solution.
...
+  must be documented (how else will people find out about your awesome work?).
 - It can't change existing syntax except in extremely well-reasoned cases. If
   you're changing the syntax of existing features, definitely open an issue
   first so we can discuss.
 
+#### Test Coverage
+
+##### Our Testing Goals
+
+We want to be sure of a few things, and testing will help us accomplish this:
```
BrianHicks: you can remove "help us accomplish this" and get the same meaning.

CONTRIBUTING.md
```diff
@@ -71,7 +95,7 @@ harder to review. Pretty please?
 ## Code of Conduct
 
 Participation in the development process of Converge is subject to
-the [code of conduct](CODE_OF_CONDUCT.md). Please familiarise yourself with that
+the [code of conduct](CODE_OF_CONDUCT.md). Please familiarize yourself with that
```
BrianHicks: :gb:

CONTRIBUTING.md
rebeccaskinner: We were talking about this a bit in the office.  I think we should avoid mocking whenw e can, but prefer mocking over having untested/untestable code (e.g. when the code is only available on one platform)- but we should probably have further discussion about this.

rebeccaskinner: I think if we're going to tell people "you shouldn't mock, except when you should, but usually not" it's probably better to just not mention it at all and leave it at the fact that we should have tests for things.

CONTRIBUTING.md
arichardet: Updated: left the period for the fragment, and removed from the complete sentences.

PR #370, opened 2016-10-12
FIX local command flag added to example

philcryer:
As discussed in Slack, just a simple fix in the getting started page. The user is running locally, so you need the `--local` flag in the example command, just like all the others are on the page, to allow it to run without error.

```
- Now try running `converge plan helloYou.hcl`. The same thing happens as if you had called the module yourself!
+ Now try running `converge plan --local helloYou.hcl`. The same thing happens as if you had called the module yourself!
```


BrianHicks:
Thanks @philcryer!

PR #372, opened 2016-10-12
use a container for CI

BrianHicks:
We had a bunch of dependencies we were installing before CI would begin. No need to do those all the time. Let's just use a custom container. Thanks Wercker!

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ runs tasks in parallel.
     - [Development](#development)
         - [Tools](#tools)
         - [RPC](#rpc)
+        - [Continuous Integration and Testing](#continuous-integration-and-testing)
     - [License](#license)
 
 <!-- markdown-toc end -->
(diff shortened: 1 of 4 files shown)


BrianHicks:
Try this on for size! `docker run -i -t --rm --volume $(pwd):/go/src/github.com/asteris-llc/converge asteris/converge-ci /bin/bash -c 'cd /go/src/github.com/asteris-llc/converge; make test'`

BrianHicks:
Should we document that?

stevendborrelli:
This is great! Please add a line to the readme and docs and we can merge it.

stevendborrelli:
Maybe after https://github.com/asteris-llc/converge/pull/364 is merged we can use the .hcl files to build ci containers for the different distributions.

BrianHicks:
Done

PR #375, opened 2016-10-14
Update user status level and errors, add checks in Apply for group

arichardet:
- Updates status level and errors for user
- Adds checks in Apply() when adding a user with a group
  - Verifies that the group name or gid exists prior to adding
  - Returns useful error if the group does not exist

### DIFF
--- a/resource/user/user.go
+++ b/resource/user/user.go
@@ -93,7 +93,7 @@ func (u *User) Check(resource.Renderer) (resource.TaskStatus, error) {
 	status := &resource.Status{}
 
 	if nameErr == ErrUnsupported {
-		status.Level = resource.StatusFatal
+		status.RaiseLevel(resource.StatusFatal)
 		return status, ErrUnsupported
 	}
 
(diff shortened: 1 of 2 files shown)


resource/user/user.go
```diff
@@ -236,16 +245,21 @@ func (u *User) Apply() (resource.TaskStatus, error) {
 
 			switch {
 			case nameNotFound:
-				options := SetAddUserOptions(u)
-				err := u.system.AddUser(u.Username, options)
+				options, err := SetAddUserOptions(u)
 				if err != nil {
-					status.Level = resource.StatusFatal
+					status.RaiseLevel(resource.StatusCantChange)
+					status.Output = append(status.Output, err.Error())
```
stevendborrelli: Should this be part of the error itself instead of in messages?

stevendborrelli: The error shows up in 2 places:
```
root/user.user.user:
 Error: root/user.user.user still has changes after apply: will not attempt add: user test
 Messages:
  group foo does not exist
 Has Changes: yes
 Changes: No changes
```

arichardet:
@stevendborrelli ready for another look - I updated all errors for user.

PR #379, opened 2016-10-14
named lock

### LINKED ISSUE #249: named locks
A named lock would be nice for doing things that affect the whole system, but which Converge currently does in parallel. (Like package management using `apk` on Alpine)

ryane:
fixes #249


docs/content/dependencies.md
```diff
@@ -142,3 +142,82 @@ would all work like [the param example in the getting started guide]({{< ref
...
+## Locking
+
+There can be some scenarios where a group of tasks are not explicitly dependent
+on each other but also cannot be run in parallel. A good example of this is
+package management tools like
+[apk](http://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management) or
+[apt](https://wiki.debian.org/Apt). As an example, let's look at this file which
+installs three packages:
+
+```hcl
```
BrianHicks: in "real life" these tasks would probably all be dependent on an `apt-get update` instead of inlining it. No reason to pull thrice in a run. Let converge deduplicate :)

docs/content/dependencies.md
```diff
@@ -142,3 +142,82 @@ would all work like [the param example in the getting started guide]({{< ref
...
+task "install-build-essential" {
+  check = "dpkg -s build-essential >/dev/null 2>&1"
+  apply = "apt-get update 2>&1 > /dev/null && apt-get -y install build-essential"
+}
+```
+
+Here is what the corresponding graph looks like:
+
+{{< figure src="/images/dependencies/without-locks.png"
+           caption="The graph output of the above module. Converge will attempt
```
BrianHicks: probably shouldn't wrap here since this will be inserted verbatim into the alt text.

docs/content/dependencies.md
```diff
@@ -142,3 +142,82 @@ would all work like [the param example in the getting started guide]({{< ref
...
+
+And the corresponding graph:
+
+{{< figure src="/images/dependencies/with-locks.png"
+           caption="The graph output of the above module. The tasks in the lock
+           will not run in parallel." >}}
+
+{{< note title="Future Improvements" >}}
+In this example, we are installing packages by calling `apt-get` in Converge
+tasks. We do plan to build higher-level resources to handle package management
```
BrianHicks: "do plan" -> "plan"

samples/locks.hcl
```diff
@@ -0,0 +1,34 @@
+task "file1" {
```
BrianHicks: do you think we could use a more realistic example here?

BrianHicks:
This also doesn't appear to work across modules. Does it? Could you also hide the locks in the human printer by default? Now might be a good time to change  "show-params" to "show-meta" as the flag there.

Here's what you get with a module-ized workflow now:

withlocks

and after removing the lock nodes:

withoutlocks

No dependencies! Uh oh!

---

apt.hcl:

```hcl
param "package" {}

task "update" {
  check = "echo always updating; exit 1"
  apply = "apt-get update"
}

task "install" {
  check   = "dpkg -s {{param `package`}} >/dev/null 2>&1"
  apply   = "apt-get install -y {{param `package`}}"
  depends = "task.update"
  lock    = "apt"
}
```

packages.hcl:

```hcl
module "apt.hcl" "jq" {
  params {
    package = "jq"
  }
}

module "apt.hcl" "tree" {
  params {
    package = "tree"
  }
}

module "apt.hcl" "build-essential" {
  params {
    package = "build-essential"
  }
}
```

ryane:
closing in favor of #392

Issue #385, opened 2016-10-17
Explicit dependencies fail inside of case statements


rebeccaskinner:
This is broken:
```
file.content "foo" {
  destination = "foo"
  content = "foo"
}

switch "touch-file" {
  default {
    task.query "touch-file" {
      query = "touch file.txt"
      depends = ["file.content.foo"]
    }
  }
}
```

Probably because dependency resolution isn't addressing ancestor nodes for explicit dependencies, we should probably look at the code that's being used by lookup and param dependency generation and either put that in getDepends or move the logic to the actual walk function.

rebeccaskinner:
Closed by #391

PR #391, opened 2016-10-18
Fix/385 dependencies in conditionals

rebeccaskinner:
fix `depends =` clause in branch nodes, and fix regression related to premature unwrapping of branch macro nodes.


stevendborrelli:
Fixes #385

ryane:
lgtm on rebase

PR #392, opened 2016-10-18
named groups

### LINKED ISSUE #249: named locks

ryane:
fixes #249
replaces #379


stevendborrelli:
Will `group` clash with settings for things like file ownership, etc?

ryane:
Yeah, I suppose it would (doesn't look like we have any fields call just "group" right now though). Should we rename it back to "lock"? something else?

rebeccaskinner:
Two nodes that share a group and also have an explicit dependency seems to cause a deadlock.  Example:

```hcl
task.query "a" {
  query = "echo a"
  group = "query"
  depends = ["task.query.b"]
}

task.query "b" {
  query = "echo b"
  group = "query"
}
```

load/dependencyresolver.go
```diff
@@ -174,3 +193,157 @@ func getNearestAncestor(g *graph.Graph, id, node string) (string, bool) {
...
+	nodes []*node.Node
+}
+
+func (b byDependencyCount) Len() int      { return len(b.nodes) }
+func (b byDependencyCount) Swap(i, j int) { b.nodes[i], b.nodes[j] = b.nodes[j], b.nodes[i] }
+func (b byDependencyCount) Less(i, j int) bool {
+	return len(b.g.Dependencies(b.nodes[i].ID)) > len(b.g.Dependencies(b.nodes[j].ID))
+}
+
+func groupDeps(ctx context.Context, g *graph.Graph, group string) (*graph.Graph, error) {
```
rebeccaskinner: Is there any way that we can reasonably break this up to factor out smaller helper functions?  Like maybe pulling some of the individual for loops out into helper functions?

ryane: I broke out each stage into separate functions.

rebeccaskinner:
LGTM!

ryane:
thx!

PR #400, opened 2016-10-19
Allow lists in parameters when called in a module

### LINKED ISSUE #397: Use lists as params in modules is broken
Given the following module placed in `samples/`:

```hcl
module "paramJoin.hcl" "paramJoin" {
  params {
    items = ["a", "b", "c"]
  }
}
```

We get the following error:

```
timestamp="2016-10-19T09:38:48-05:00" level="ERROR" msg="could not render" component="rpc" error="error rendering field content: template: content:1:2: executing \\\"content\\\" at <paramList `items`>: error calling paramList: param is not a list, it is a string ([a b c])" location="samples/sourceFileParamJoin.hcl" runID="1e64bbb5-8b75-4992-9b23-4bfcbb0c374b"
timestamp="2016-10-19T09:38:48-05:00" level="FATAL" msg="could not get responses" component="client" error="error getting status response: rpc error: code = 2 desc = rendering samples/sourceFileParamJoin.hcl: error rendering field content: template: content:1:2: executing \\\"content\\\" at <paramList `items`>: error calling paramList: param is not a list, it is a string ([a b c])" file="samples/sourceFileParamJoin.hcl"
```

BrianHicks:
This allows literal lists in parameters when calling a module.

Fixes #397

### DIFF
--- a/helpers/fakerenderer/fakerenderer.go
+++ b/helpers/fakerenderer/fakerenderer.go
@@ -14,10 +14,12 @@
 
 package fakerenderer
 
+import "github.com/asteris-llc/converge/resource"
+
 // FakeRenderer is a pass-through renderer for testing resources
 type FakeRenderer struct {
 	ID           string
-	DotValue     string
+	DotValue     resource.Value
 	ValuePresent bool
 }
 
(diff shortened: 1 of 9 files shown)


helpers/fakerenderer/fakerenderer.go
```diff
@@ -27,7 +29,7 @@ func (fr *FakeRenderer) GetID() string {
 }
 
 // Value returns a blank string
```
rebeccaskinner: Not really true anymore since we return a resource.Value, and it's not necessarily empty.

PR #403, opened 2016-10-20
Handle thunked branches

rebeccaskinner:
Preliminary fix - I'm doing more thorough testing now but okay for review.


blackbox/test_conditional.sh
```diff
@@ -0,0 +1,118 @@
...
+fi
+
+if [ ! -f foo-file-2.txt ]; then
+    echo "foo-file-2.txt doesn't exist!"
+    exit 1
+fi
+
+if [[ "$(cat foo-file.txt)" != "foo1" ]]; then
+    echo "foo-file.txt doesn't have the right content"
+    echo "has '$(at foo-file.txt)', want 'foo1'"
```
BrianHicks: I think you want "cat" here and throughout similar statements?

blackbox/test_conditional.sh
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env bash
```
BrianHicks: this seems like two tests in one? Is it testing two distinct modes?

render/preprocessor/preprocessor.go
```diff
@@ -257,82 +257,130 @@ func HasMethod(obj interface{}, methodName string) bool {
...
+func FieldMap(obj interface{}) map[string]reflect.Value {
+	embeddedRefs := make(map[string]int)
+	unembeddedRefs := make(map[string]int)
+
+	results := make(map[string]reflect.Value)
+	val := reflect.ValueOf(obj)
+	if canBeNil(val) && val.IsNil() {
+		return results
+	}
+	for ; (val.Kind() == reflect.Ptr && !val.IsNil()) || val.Kind() == reflect.Interface; val = val.Elem() {
```
BrianHicks: why is this for block empty?

transform/control.go
```diff
@@ -50,6 +52,30 @@ func ResolveConditionals(ctx context.Context, g *graph.Graph) (*graph.Graph, err
 				if !ok {
 					continue
 				}
+
+				if childThunk, ok := targetPreparerMeta.Value().(*render.PrepareThunk); ok {
+					junk := make([]byte, 32)
```
BrianHicks: what's this all about?

blackbox/test_conditional.sh
rebeccaskinner: We can do that, I just thought that it would make sense to have one general test for all things conditional.  What about leaving it as one script with separate functions?

blackbox/test_conditional.sh
rebeccaskinner: ...yep

render/preprocessor/preprocessor.go
rebeccaskinner: We're mutating val here so that as long as it's a non-nil pointer or an interface we get the inner Elem.  There's nothing to do in the block itself since all the mutation is done in the for.

blackbox/test_conditional.sh
BrianHicks: Let's say functions. Kinda like `t.Run` :P

render/preprocessor/preprocessor.go
BrianHicks: do you reckon you could rewrite it more in the usual style?

```go
for {
  if cond {
    break
  }
  // work
}
```

transform/control.go
BrianHicks: ah, that makes sense. Would you mind adding a comment to that effect?

PR #413, opened 2016-10-24
Nightly/test builds

BrianHicks:
This configuration will:

1. tag all releases made with `make converge` with the correct version metadata
2. do the same, but for CI
3. upload those releases to S3 when built on master (I've already added the pipeline steps in Wercker)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,8 +94,8 @@ vendor-clean: ${NOVENDOR}
 xcompile: rpc/pb/root.pb.go rpc/pb/root.pb.gw.go test
 	@echo "package cmd" > cmd/version_number.go
 	@echo >> cmd/version_number.go
-	@echo "const Version = \"$(shell git describe --dirty)\"" >> cmd/version_number.go
-	@echo "set version to $(shell git describe --dirty)"
+	@echo "const Version = \"$(shell git describe)\"" >> cmd/version_number.go
+	@echo "set version to $(shell git describe)"
 
 	@rm -rf build/
 	@mkdir -p build/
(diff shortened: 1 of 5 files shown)


BrianHicks:
Feel free to LGTM but please do not merge yet. Since the bucket is still being created, it will fail.

BrianHicks:
This is now unblocked.

Makefile
```diff
@@ -1,13 +1,17 @@
...
 RPCLINT=$(shell find ./rpc -type f \( -not -iname 'root.*.go' -iname '*.go' \) )
 TOLINT = $(shell find . -type f \( -not -ipath './vendor*'  -not -ipath './docs*' -not -ipath './rpc*' -not -iname 'main.go' -iname '*.go' \) -exec dirname {} \; | sort -u)
 TESTDIRS = $(shell find . -name '*_test.go' -exec dirname \{\} \; | grep -v vendor | uniq)
 NONVENDOR = ${shell find . -name '*.go' | grep -v vendor}
 BENCHDIRS= $(shell find . -name '*_test.go' | grep -v vendor | xargs grep '*testing.B' | cut -d: -f1 | xargs -n1 dirname | uniq)
 BENCH = .
 
-converge: $(shell find . -name '*.go') rpc/pb/root.pb.go rpc/pb/root.pb.gw.go
+converge: $(shell find . -name '*.go' -not -name 'version_number.go') rpc/pb/root.pb.go rpc/pb/root.pb.gw.go
+	@echo "package cmd" > cmd/version_number.go
```
rebeccaskinner: We could make this a phony target and depend on it for the converge target just to keep version number generation separate from building, but it's probably not that big of a deal.

PR #418, opened 2016-10-25
 use ldflags to set version

stevendborrelli:
The current version that creates a .go file interferes with `go get` on our repository. 

This version uses ldflags at build time, and will display `unset` if converge is built outside of
the make process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,6 @@ samples/**/*.png
 
 # release
 build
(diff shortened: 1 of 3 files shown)


BrianHicks:
LGTM otherwise. Thanks for the technique!

stevendborrelli:
.gitignore updated

PR #426, opened 2016-10-25
Fix/415 parent edges

rebeccaskinner:
This builds on top of #423 and prevents `SafeDisconnect` from removing parent edges so that conditionals are not moved outside of their branch nodes when grouped.


rebeccaskinner:
This isn't going to work as-is and breaks some stuff.

Issue #389, opened 2016-10-17
resolving conditional macros line


BrianHicks:
we get this line:

```
2016-10-17T16:05:37-05:00 |INFO| resolving conditional macros	component=rpc function=ResolveConditionals runID=4e5c5f77-ccfc-4c11-85e6-5a7a08c8a65c
```

on every traversal of the graph. This needs to go away or be lowered to debug

BrianHicks:
Fixed by #428. Thanks @mason-fish!

PR #428, opened 2016-10-26
[389] switch log level from Info to Debug

mason-fish:
in reference to [issue #389](https://github.com/asteris-llc/converge/issues/389)


BrianHicks:
Thanks @mason-fish!

PR #434, opened 2016-10-26
Add the ability to modify a user

### LINKED ISSUE #425: Error when adding a user when a group of the same name exists
If adding a user "alice", and a group "alice" exists:
- The plan indicates the user will be added
- The apply then fails with an unhelpful error message from the useradd command: "useradd: exit status 9"

arichardet:
Add the ability to modify a user #278 
Fixes #425

### DIFF
--- a/docs/content/resources/user.group.md
+++ b/docs/content/resources/user.group.md
@@ -1,7 +1,7 @@
 ---
 title: "user.group"
 slug: "user-group"
-date: "2016-10-04T13:01:50-05:00"
+date: "2016-10-28T08:47:08-05:00"
 menu:
   main:
     parent: resources
@@ -43,5 +43,6 @@ The group Name will be changed to NewName.
   Valid values: `present` and `absent`
 
   State is whether the group should be present.
+The default value is present.
 
 
--- a/docs/content/resources/user.user.md
+++ b/docs/content/resources/user.user.md
@@ -1,7 +1,7 @@
 ---
 title: "user.user"
 slug: "user-user"
-date: "2016-10-04T13:01:50-05:00"
+date: "2016-10-28T08:47:08-05:00"
 menu:
   main:
     parent: resources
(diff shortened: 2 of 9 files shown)


rebeccaskinner:
Could you add a `samples/userModify.hcl` (and a blackbox test if you think we need one)?  I've noticed that we've had a lot of regressions introduced lately that we could have caught earlier if we'd had more cases being exercised by blackbox tests.  Thanks!

rebeccaskinner:
Using this module as an example:
```
user.user "created" {
  username = "test1"
  home_dir  = "/home/test1-home"
  state    = "present"
}

file.content "a" {
  destination = "/home/test1-home/foo"
  content = "foo"
  depends = ["user.user.created"]
}

user.user "modified" {
  username = "test1"
  home_dir   = "/home/test1"
  move_dir  = true
  groupname = "daemon"
  depends   = ["user.user.created","file.content.a"]
}
```

I found it a bit surprising that during initial user creation the home directory wasn't created, but then during user modification if the new home directory target existed it would give me an error.

It would be nice if it would create the home directory for you (unless you specify not to) and if it would not error when moving the home directory (maybe just give the user ownership over everything in that directory and merge it with the existing home directory).

If merging the directories isn't really feasible it would be nice to have a more explicit error message, the one I got was:

```
root/user.user.modified:
 Error: user modify: usermod: exit status 12
 Messages:
 Has Changes: yes
 Changes: No changes
```

arichardet:
I added fields for adding a user:
- `create_home` creates the home directory
- `skel_dir` contents are copied to the home directory if `create_home` is true

arichardet:
The error returned when modifying a user `home_dir` is "usermod: exit status 12", which indicates the `home_dir` specified for the modification already exists. This is not an error and the home directory name is modified to the new value indicated. Since we are unsure what other corner cases may be encountered with this status code, I am not in favor of suppressing the error.

PR #447, opened 2016-10-31
Fix/graceful exit

### LINKED ISSUE #319: resources should have the ability to gracefully stop on interrupt
The Task interface should be extended to take a Context so that resources can gracefully stop (and cleanup, if necessary) when a converge run is interrupted.

arichardet:
Add context to allow for graceful exit from resources upon interrupt.
Fixes #319


arichardet:
It sounds like there is agreement to merge and open issues for the existing resources.

Issue #325, opened 2016-09-29
Improvement in difference engine


stevendborrelli:
1. I propose adding a new method to status.go:
```
// Difference returns the Current value of a Difference by key
func (t *Status) Difference(key string) (string, bool) {
	diff, ok := t.Differences[key]
	if !ok {
		return "", ok
	}
	return diff.Current(), diff.Changes()
}
```

Usage would be like this in an `Apply()` phase to check if file content has been modified:
```
	if content, ok := status.Difference("content"); ok && f.Type == "file" {
		err = ioutil.WriteFile(f.Destination, content, f.FileMode)
```

2. As part of this, the engine and Check() should use the `status.go` `status.AnyChanges()` method. 

This means that a typical check would just `AddDifference()` for every field, and `AnyChanges()`  is used to see if any of the defined differences indicate a change. 

Instead of doing this in my Check() method:
```
	if f.State != actual.State {
		status.AddDifference("state", actual.State, f.State, "")
	}
```

I'd just do:
```
status.AddDifference("state", actual.State, f.State, "")

```

And expect anything else to call `AnyChanges()` for any change or `Difference()` to get the value of a specific change.

ryane:
👍  The docker container resource already follows the pattern of calling `status.AddDifference` without comparing first which mostly works today. But, other parts of the system need updating. For example, the printer doesn't check if the diffs have changes so you end up with output like described in https://github.com/asteris-llc/converge/pull/317#issuecomment-249736326.

PR #453, opened 2016-10-31
Feature/docker.volume

### LINKED ISSUE #299: docker.volume resource
Ability to manage docker data volumes: https://docs.docker.com/engine/reference/commandline/volume_create/

ryane:
fixes #299


stevendborrelli:
Getting a panic when I try to run the sample:
```
./converge plan --local samples/dockerVolume.hcl
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x421382]

goroutine 22 [running]:
panic(0x6308e0, 0xc4200120a0)
    /usr/local/Cellar/go/1.7.1/libexec/src/runtime/panic.go:500 +0x1a1
github.com/asteris-llc/converge/resource/docker/volume.(*Volume).Check(0xc420175080, 0xa10100, 0xc42015d900, 0x20, 0xc42015d900, 0x0, 0x0)
    /Users/steve/go/src/github.com/asteris-llc/converge/resource/docker/volume/volume.go:65 +0x3c2
```

ryane:
panic should be fixed + rebased

resource/docker/volume/volume.go
```diff
@@ -0,0 +1,165 @@
...
+		v.Status.Level = resource.StatusFatal
+		return v, err
+	}
+
+	v.Status.AddDifference(v.Name, string(volumeState(vol)), string(v.State), "")
+
+	if v.State == StatePresent && vol != nil && v.Force {
+		expectedLabels := mapToString(v.Labels)
+		actualLabels := mapToString(vol.Labels)
+		v.Status.AddDifference("labels", actualLabels, expectedLabels, "")
```
stevendborrelli: Shouldn't we only add these diffs if they don't match?

ryane: it's an option but this approach is consistent with the existing docker resources. this is discussed more in #325. let me know if you think I should change it.

we could also update the human printer to only print diffs with changes. something like:

```diff
diff --git a/prettyprinters/human/human.go b/prettyprinters/human/human.go
index 74d41ac..4e137a8 100644
--- a/prettyprinters/human/human.go
+++ b/prettyprinters/human/human.go
@@ -136,7 +136,9 @@ func (p *Printer) DrawNode(g *graph.Graph, id string) (pp.Renderable, error) {
 	Has Changes: {{if .HasChanges}}{{yellow "yes"}}{{else}}no{{end}}
 	Changes:
 		{{- range $key, $values := .Changes}}
+		{{- if $values.Changes}}
 		{{cyan $key}}:	{{diff ($values.Original) ($values.Current)}}
+		{{- end}}
 		{{- else}} No changes {{- end}}
 
 `)
```

Issue #446, opened 2016-10-28
package.rpm state does not default to present


ryane:
example:

``` hcl
package.rpm "mc" {
  name = "mc"
}
```

```
$ /usr/local/bin/converge plan --local rpm.hcl
2016-10-28T15:00:56Z |INFO| serving     addr=:47740 component=rpc
2016-10-28T15:00:56Z |WARN| skipping module verification        component=client
2016-10-28T15:00:56Z |INFO| got status  component=client file=rpm.hcl id=root/package.rpm.mc run=STARTED stage=PLAN
2016-10-28T15:00:57Z |INFO| got status  component=client file=rpm.hcl id=root run=STARTED stage=PLAN

root/package.rpm.mc:
 Messages:
 Has Changes: yes
 Changes:
  mc: "absent" => ""

Summary: 0 errors, 1 changes

$ /usr/local/bin/converge apply --local rpm.hcl
2016-10-28T15:01:03Z |INFO| serving     addr=:47740 component=rpc
2016-10-28T15:01:03Z |WARN| skipping module verification        component=client
2016-10-28T15:01:03Z |INFO| got status  component=client file=rpm.hcl id=root/package.rpm.mc run=STARTED stage=APPLY
2016-10-28T15:01:04Z |INFO| got status  component=client file=rpm.hcl id=root run=STARTED stage=APPLY

root/package.rpm.mc:
 Error: root/package.rpm.mc still has changes after apply: exit status 1
 Messages:
 Has Changes: yes
 Changes:
  mc: "absent" => ""

Summary: 2 errors, 1 changes

 * root/package.rpm.mc: root/package.rpm.mc still has changes after apply: exit status 1
 * root: error in dependency "root/package.rpm.mc"
```

The preparer uses a `default` struct tag but it looks like that tag is not actually implemented. We should either explicitly require `State` to be specified or default it to "present".

If we do the latter, should we implement a `default` struct tag?

stevendborrelli:
Feels like having a default struct tag would make things easier for module writers.

rebeccaskinner:
fixed in #463

PR #463, opened 2016-11-02
default to state 'present' for rpm module

rebeccaskinner:


ryane:
LGTM on green (looks like missing "r" on the test func comment)

rebeccaskinner:
fixes #446

PR #477, opened 2016-11-04
docker.network resource

### LINKED ISSUE #298: docker.network resource
Support for creating docker [networks](https://docs.docker.com/engine/userguide/networking/). Currently with Converge, you can only configure communication between containers using the legacy links feature.

ryane:
fixes #298

### DIFF
--- a/resource/docker/network/network.go
+++ b/resource/docker/network/network.go
@@ -37,6 +37,9 @@ const (
 
 	// DefaultDriver is the default network driver
 	DefaultDriver = "bridge"
+
+	// DefaultIPAMDriver is the default IPAM driver
+	DefaultIPAMDriver = "default"
 )
 
 // Network is responsible for managing docker volumes
@@ -48,6 +51,7 @@ type Network struct {
 	Driver  string
 	Labels  map[string]string
 	Options map[string]interface{}
+	IPAM    dc.IPAMOptions
 	State   State
 	Force   bool
 }
@@ -65,9 +69,7 @@ func (n *Network) Check(resource.Renderer) (resource.TaskStatus, error) {
 	n.AddDifference(n.Name, string(networkState(nw)), string(n.State), "")
 
 	if n.State == StatePresent && nw != nil && n.Force {
-		n.AddDifference("labels", mapCompareStr(nw.Labels), mapCompareStr(n.Labels), "")
-		n.AddDifference("driver", nw.Driver, n.Driver, DefaultDriver)
-		n.AddDifference("options", mapCompareStr(nw.Options), mapCompareStr(toStrMap(n.Options)), "")
+		n.diffNetwork(nw)
 	}
 
 	if resource.AnyChanges(n.Differences) {
@@ -111,7 +113,9 @@ func (n *Network) Apply() (resource.TaskStatus, error) {
 			Driver:  n.Driver,
 			Labels:  n.Labels,
 			Options: n.Options,
+			IPAM:    n.IPAM,
 		}
+
 		nw, err = n.client.CreateNetwork(opts)
 		if err != nil {
 			n.RaiseLevel(resource.StatusFatal)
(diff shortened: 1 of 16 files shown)


resource/docker/network/preparer_test.go
```diff
@@ -0,0 +1,62 @@
...
+// TestPreparerInterface ensures that the correct interfaces are implemented by
+// the preparer
+func TestPreparerInterface(t *testing.T) {
+	t.Parallel()
+	assert.Implements(t, (*resource.Resource)(nil), new(network.Preparer))
+}
+
+// TestPreparerPrepare tests the Prepare function
+func TestPreparerPrepare(t *testing.T) {
+	t.Run("name is required", func(t *testing.T) {
```
arichardet: can remove this test with change in network/preparer.go (handled in resource/preparer.go)

resource/docker/network/preparer.go
```diff
@@ -0,0 +1,163 @@
...
+	if p.Driver == "" {
+		p.Driver = DefaultDriver
+	}
+
+	if p.IPAMDriver == "" {
+		p.IPAMDriver = DefaultIPAMDriver
+	}
+
+	if p.State == "" {
+		p.State = "present"
```
arichardet: could use StatePresent instead of "present"

resource/docker/network/network_test.go
```diff
@@ -0,0 +1,406 @@
...
+			nw := &network.Network{Name: nwName, State: "absent"}
+			c := &mockClient{}
+			nw.SetClient(c)
+			c.On("FindNetwork", nwName).Return(&dc.Network{Name: nwName}, nil)
+
+			status, err := nw.Check(context.Background(), fakerenderer.New())
+			require.NoError(t, err)
+			assert.True(t, status.HasChanges())
+			assert.Equal(t, 1, len(status.Diffs()))
+			comparison.AssertDiff(t, status.Diffs(), nwName, "present", "absent")
```
arichardet: could use StatePresent and StateAbsent throughout

resource/docker/network/preparer.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// indicates whether or not the volume will be recreated if the state is not
+	// what is expected. By default, the module will only check to see if the
+	// volume exists. Specified as a boolean value
+	Force bool `hcl:"force"`
+}
+
+// Prepare a docker network
+func (p *Preparer) Prepare(ctx context.Context, render resource.Renderer) (resource.Task, error) {
+	if p.Name == "" {
```
ryane: I think this is needed to prevent someone from manually specifying `name = ""` in the hcl

resource/docker/network/network.go
```diff
@@ -0,0 +1,240 @@
...
+	)
+	sort.Strings(pairs)
+	return strings.Join(pairs, ", ")
+}
+
+// IPAMConfigs is a slice of dc.IPAMConfig
+type IPAMConfigs []dc.IPAMConfig
+
+// Len implements the sort interface for IPAMConfigs
+func (ic IPAMConfigs) Len() int { return len(ic) }
```
ryane: these are used to implement the sort interface: https://golang.org/pkg/sort/

resource/docker/network/network_test.go
ryane: true but it requires a cast (otherwise, I get `cannot use network.StatePresent (type network.State) as type string in argument to comparison.AssertDiff`). makes the lines really long and harder to read (imo). worth it?

resource/docker/network/preparer_test.go
ryane: see https://github.com/asteris-llc/converge/pull/477#discussion_r87093311

resource/docker/network/network_test.go
ryane: actually wasn't so bad - went ahead and fixed these

resource/docker/network/preparer.go
arichardet: yes, that is correct

resource/docker/network/network.go
arichardet: I see

resource/docker/network/preparer_test.go
arichardet: yes, this is ok

arichardet:
LGTM!

PR #484, opened 2016-11-07
Fix/conditional nodes

### LINKED ISSUE #386: Case predicates fail when they include `lookup`
Cases that contain `lookup` in the predicates fail because we don't do proper handling of `render.PrepareThunk` types in `transform.Control`.  This should be pretty straightfoward to fix, but right now control transformation is initiated by render so it creates an import loop, we probably need to slightly reorganize the code then do a little bit of introspection.

rebeccaskinner:
Fixes #386 

Posted for initial review:  May require additional dead-code removal.


graph/merge_duplicates.go
```diff
@@ -59,6 +59,7 @@ func MergeDuplicates(ctx context.Context, g *Graph, skip SkipMergeFunc) (*Graph,
 		// if we haven't seen this value before, register it and return
 		target, ok := values[hash]
 		if !ok {
+
```
ryane: why all of these extra new lines?

rebeccaskinner: regex to remove debugging statements didn't remove the lines... fixing

PR #497, opened 2016-11-14
release notes

rebeccaskinner:
don't merge until we're ready to release


docs/content/release-notes/0.4.0.md
```diff
@@ -0,0 +1,161 @@
...
+    // Username is the user login name.
+    Username string `hcl:"username" required:"true" nonempty:"true"`
+```
+
+when `nonempty:"true"` is set, an error will be returned if the user sets the
+value of a field to the zero value for it's type.
+
+#### Time Duration Types
+
+Resource structs may now use `time.Duration` as a field type.  Duration strings
```
arichardet: In addition to a duration string, a number in seconds is acceptable input.

PR #509, opened 2016-11-17
move rendering plant creation inside of the transform block to avoid …

rebeccaskinner:
…potential race conditions


arichardet:
LGTM

PR #512, opened 2016-11-18
docker: update docker dependency

### LINKED ISSUE #485: docker.network misleading error messages
currently, the docker.network can return misleading error messages. if there is any scenario where creating a network results in the docker api returning a 409 (conflict) status, it will report that the network could not be created because it already exists. However, this is not always the actual reason for the error. For example, if you attempt to create a network that has overlapping ip addresses with an existing network, you will see the error "network already exists" instead of "networks have overlapping IPv4". This is due to an issue in the underlying docker api client library: https://github.com/fsouza/go-dockerclient/issues/605

When the underlying issue is fixed, we can update the dependency and the correct error message should be returned.

ryane:
fixes #485 

see https://github.com/fsouza/go-dockerclient/issues/605

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -49,7 +49,7 @@ imports:
 - name: github.com/fsnotify/fsnotify
   version: 629574ca2a5df945712d3079857300b5e4da0236
 - name: github.com/fsouza/go-dockerclient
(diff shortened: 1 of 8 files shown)


arichardet:
LGTM, please rebase.

PR #527, opened 2016-11-23
Fix/disable docker solaris

ryane:


arichardet:
A couple things:
- It is probably a good idea to return an error that docker is not supported if converge is run on solaris.
- You could add a comment to the hcl samples that docker is not supported on solaris; this would be in the docs then too.

ryane:
thx, good idea, I added a comment to the hcl samples. 

But, on the first suggestion, I'm not totally sure where to return an error. right now, the docker resources are not compiled into the solaris build at all so I think it should be like they don't exist. They are not registered in the registry. While I don't have a way to actually verify on solaris, I believe you should get an error like:

```shell
$ ./converge plan --local samples/dockerImage.hcl
2016-11-23T17:49:39-05:00 |INFO| serving        addr=:47740 component=rpc
2016-11-23T17:49:39-05:00 |WARN| skipping module verification   component=client
2016-11-23T17:49:39-05:00 |ERROR| could not load        component=rpc error=could not resolve resources: 1 error(s) occurred:

* root/docker.image.busybox: "docker.image" is not a valid resource type in "docker.image.busybox" location=samples/dockerImage.hcl runID=d8ff30e7-6a2d-471d-aa8e-f1e6506dee76
2016-11-23T17:49:39-05:00 |FATAL| could not get responses       component=client error=error getting status response: rpc error: code = 2 desc = loading samples/dockerImage.hcl: could not resolve resources: 1 error(s) occurred:

* root/docker.image.busybox: "docker.image" is not a valid resource type in "docker.image.busybox" file=samples/dockerImage.hcl
```

if you were to use `docker.image`, for example. Do we need more than this? I suppose we could maintain a map of unsupported resources by platform in the `load` package and return a more specific error.

PR #533, opened 2016-12-01
Fix/374 immutable status

rebeccaskinner:


executor/pipeline.go
```diff
@@ -38,6 +38,17 @@ func (p Pipeline) AndThen(f PipelineFunc) Pipeline {
 	return p
 }
 
+// LogAndThen is a utility function for logging purposes, it logs calls to
+// AndThen with printf
+func (p Pipeline) LogAndThen(msg string, f PipelineFunc) Pipeline {
```
BrianHicks: msg isn't used here at all. What's it for?

render/renderer.go
```diff
@@ -196,30 +201,48 @@ func (r *Renderer) lookup(name string) (string, error) {
 	}
 
 	if _, isThunk := meta.Value().(*PrepareThunk); isThunk {
-		log.WithField("proxy-reference", vertexName).Warn("node is unresolvable")
+		log.WithField("proxy-reference",
```
BrianHicks: this and the Sprintf call below are nested a little weird. Could you change it to the…

```go
x(
    a,
    b,
).c(
    d(
        e,
        f,
    )
)
```

… style?

render/renderer.go
```diff
@@ -196,30 +201,48 @@ func (r *Renderer) lookup(name string) (string, error) {
...
+		).Warn(fmt.Sprintf("%s: cannot resolve %s in node %s in prepare thunk",
+			r.ID,
+			vertexName, terms,
+		))
 		r.resolverErr = true
 		return "", ErrUnresolvable{}
 	}
 
 	if _, isPreparer := meta.Value().(*resource.Preparer); isPreparer {
 		log.WithField("preparer-reference", vertexName).Warn("node is unresolvable")
```
BrianHicks: probably don't need two log lines saying the same thing

resource/docker/container/container.go
```diff
@@ -47,23 +47,23 @@ var (
 type Container struct {
 	*resource.Status
```
BrianHicks: these and below are still embedding status. Wasn't part of the goal to get away from that?

resource/field_extractor.go
```diff
@@ -0,0 +1,278 @@
...
+	it's embedded in, both will be exported with the embedded struct being
+	accessible with 'StructName.FieldName' */
+
+var (
+	// ErrNilStruct is returned if there is an attempt to introspect a nil value
+	ErrNilStruct = errors.New("cannot inspect types with a nil value")
+
+	// ErrNonStructIntrospection is returned when attempting to introspect fields
+	// on a non-struct type.
+	ErrNonStructIntrospection = errors.New("cannot inspect non-struct fields")
```
BrianHicks: This reads as "cannot inspect (non-struct fields)" instead of "cannot inspect non-struct (fields)" if that makes sense… It sounds like the error is complaining that the field content is not a struct

docs/extract.go
```diff
@@ -123,7 +137,38 @@ func main() {
...
+			fset,
+			taskPath,
+			nil,
+			parser.ParseComments,
+		)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		fset = token.NewFileSet()
```
arichardet: It looks like this repeats the same 10 lines above?

docs/content/resource-authors-guide.md
```diff
@@ -134,10 +135,39 @@ type Shell struct {
...
+the following task:
+
+```hcl
+task "foo" {
+  check = "test -f foo.txt"
+  apply = "touch foo.txt"
+}
+```
+
+we may reference any of the exported fields in a lookup by typing `"{{lookup
```
arichardet: Capitalize beginning of sentence

docs/content/resource-authors-guide.md
```diff
@@ -134,10 +135,39 @@ type Shell struct {
...
+    ResultsContext
+    ExitStatus uint32 `export:"exitstatus"`
+    Stdout     string `export:"stdout"`
+    Stderr     string `export:"stderr"`
+    Stdin      string `export:"stdin"`
+    State      *os.ProcessState
+}
+```
+
+since `CommandResults` exports `stdout`, `stderr`, and `stdin` we can reference
```
arichardet: Capitalize here too

docs/content/resource-authors-guide.md
```diff
@@ -134,10 +135,39 @@ type Shell struct {
...
+task "foo" {
+  check = "test -f foo.txt"
+  apply = "touch foo.txt"
+}
+```
+
+we may reference any of the exported fields in a lookup by typing `"{{lookup
+'task.foo.field'}}"`; for example `"{{lookup 'task.foo.dir'}}"` or `"{{lookup
+'task.foo.env'}}"`.  Re-exported fields will provide a namespace for structs
+that also export values.  In our `Shell` example we are re-exporting a
```
arichardet: Should you mention earlier that this is a shell example, maybe before the sample hcl?

docs/content/resource-authors-guide.md
```diff
@@ -134,10 +135,39 @@ type Shell struct {
...
+    Stderr     string `export:"stderr"`
+    Stdin      string `export:"stdin"`
+    State      *os.ProcessState
+}
+```
+
+since `CommandResults` exports `stdout`, `stderr`, and `stdin` we can reference
+these values, for example: `"{{lookup 'task.foo.status.stdout'}}"` or
+`"{{lookup 'task.foo.status.stderr'}}"`.
+
```
arichardet: I wonder about including an hcl example that uses lookups for export and re-export-as?

docs/Makefile
```diff
@@ -34,12 +34,14 @@ resources.mk: sources.csv resourceContent.mk.tmpl
 	@echo >> $@
 	@for line in $$(cat sources.csv); do \
     sed -e "s|@@NAME|$$(echo $$line | cut -d, -f1 | tr -d '\n')|g" \
-        -e "s|@@SLUG|$$(echo $$line | cut -d, -f1 | tr -d '\n' | sed 's/\./-/g')|g" \
-        -e "s|@@SOURCE|$$(echo $$line | cut -d, -f2 | tr -d '\n')|g" \
-        -e "s|@@EXAMPLE|$$(echo $$line | cut -d, -f3 | tr -d '\n')|g" \
-        -e "s|@@TYPE|$$(echo $$line | cut -d, -f4 | tr -d '\n')|g" \
-        -e "s|@@TARGET|content/resources/$$(echo $$line | cut -d, -f1 | tr -d '\n').md|g" \
-        resourceContent.mk.tmpl >> $@; \
+	-e "s|@@SLUG|$$(echo $$line | cut -d, -f1 | tr -d '\n' | sed 's/\./-/g')|g" \
```
rebeccaskinner: The spaces were replaced with tabs for consistency with other parts of the makefile (and my editor having save-hooks that does it automatically for makefiles)

docs/content/resource-authors-guide.md
arichardet: Maybe mention the Shell struct/resource for clarity/a reference, something like: "We may reference any of the fields exported by the Shell..."

docs/content/resource-authors-guide.md
```diff
@@ -110,10 +110,11 @@ your code as possible.
...
 The
-[`resource.Task`](https://godoc.org/github.com/asteris-llc/converge/resource#Task)
-interface is what you will implement to have converge run your `Check` and
-`Apply` methods.  Your `resource.Task` implementation is also what's used to
-define lookup methods from within converge.  An example of a
+[`resource.Task`](https://godoc.org/github.com/asteris-llc/converge/resource#Task) interface
+is what you will implement to have converge run your `Check` and `Apply`
+methods.  The `export` and `re-export-as` tags in your `resource.Task`
+implementation are used to define lookup methods from within converge.  An
+example of a resource is:
```
arichardet: Maybe change to "We will use the Shell resource in the following examples:" or something like that. This will also address my comment below on mentioning Shell earlier (line 152)

docs/content/resource-authors-guide.md
```diff
@@ -134,10 +135,39 @@ type Shell struct {
...
+}
+```
+
+since `CommandResults` exports `stdout`, `stderr`, and `stdin` we can reference
+these values, for example: `"{{lookup 'task.foo.status.stdout'}}"` or
+`"{{lookup 'task.foo.status.stderr'}}"`.
+
+#### Semantics of Exported Fields
 
 1. Fields that are tagged with `export` will be exported.
```
arichardet: Missed my comment on consistency of periods in this list.

apply/pipeline_test.go
```diff
@@ -0,0 +1,62 @@
...
+	meta, _ := g.Get("root/a")
+	result, err := p.Exec(context.Background(), meta.Value())
+	require.NoError(t, err)
+	asResult, ok := result.(*apply.Result)
+	require.True(t, ok)
+	_, ok = asResult.Task.(*faketask.FakeTask)
+	assert.True(t, ok)
+	applyStatus, ok := asResult.Status.(*resource.Status)
+	if !ok {
+		t.Logf("unable to convert result status to a resource status, it's type is :: %T\n", asResult.Status)
```
BrianHicks: let's leave this, but as a note for the future: we've been doing this with `assert.True(t, ok, "msg %s", sub, sub2, etc)`

docs/content/resource-authors-guide.md
```diff
@@ -94,12 +101,108 @@ presented to the user. You can also use the common Go patterm of returning `nil,
...
+```go
+type Shell struct {
+    CmdGenerator   CommandExecutor
+    CheckStmt      string                 `export:"check"`
+    ApplyStmt      string                 `export:"apply"`
+    Dir            string                 `export:"dir"`
+    Env            []string               `export:"env"`
+    Status         *CommandResults        `re-export-as:"status"`
+    CheckStatus    *CommandResults        `export:"checkstatus"`
+    HealthStatus   *resource.HealthStatus `export:"healthstatus"`
```
BrianHicks: why do we have two statuses for Shell?

docs/content/resources/filesystem.md
```diff
@@ -43,33 +43,44 @@ filesystem "mnt-me" {
...
   WantedBy is a list of dependencies, to pass to systemd .mount unit
 
+
 - `before` (list of strings)
 
   Before is a list of dependencies, to pass to systemd .mount unit
 
 
+
+
```
BrianHicks: does filesystem not expose anything?

docs/extract.go
```diff
@@ -74,22 +76,42 @@ var (
...
 {{ if .MutuallyExclusive}}
-  Only one of {{codeCommaJoin .MutuallyExclusive "or"}} may be set.
+	Only one of {{codeCommaJoin .MutuallyExclusive "or"}} may be set.
 
 {{end}}{{ if .ValidValues}}
-  Valid values: {{codeCommaJoin .ValidValues "and"}}
+	Valid values: {{codeCommaJoin .ValidValues "and"}}
+
+{{end}}{{if ne .Doc ""}}  {{.Doc}}{{end}}
+{{end}}
```
BrianHicks: this is where that extra line in every doc file is coming from. Would you mind replacing this with `{{- end -}}` to remove the extra whitespace?

docs/extract.go
```diff
@@ -74,22 +76,42 @@ var (
...
+{{ if .HasExportedFields }}
```
BrianHicks: This could probably also be a `{{-`

docs/extract.go
```diff
@@ -74,22 +76,42 @@ var (
...
+
+## Exported Fields
 
-{{end}}{{if ne .Doc ""}}  {{.Doc}}{{end}}{{end}}
+Here are the fields that are exported for use with 'lookup'.  Re-exported fields
+will have their own fields exported under the re-exported namespace.
+
+{{- range .GetExported}}
```
BrianHicks: ... and this one will mean that there's no line break between the header and the exported docs. I can see that causing problems in certain markdown viewers (since empty lines specify paragraph breaks.) Maybe remove the `{{-` here?

docs/content/resource-authors-guide.md
rebeccaskinner: Check status was so that the user would be able to access the results of `check` even during `apply`.  Otherwise you might get non-determinism if the result `check` and `apply` have different output.

PR #534, opened 2016-12-02
Use a single port

### LINKED ISSUE #286: We're using too many ports
We're assigned port 4774 right now. We were using 2693/2694 before. We could probably mux on 4774 for HTTP or RPC (by looking at the session establishment headers.) We should probably do that, and then we'd only have one port open for everything.

BrianHicks:
This PR removes the use of any ports but 4774 (server) and 47740 (local server). It maintains SSL and adds a test to ensure that we don't fall back to weaker SSL.

Fixes #286


arichardet:
Please address the codeclimate issue, otherwise LGTM.

BrianHicks:
I don't think there's really anything to be done about that TODO. It's left in place as a hint if it ever breaks as part of an upgrade.

PR #535, opened 2016-12-02
disallow some characters as resource names

### LINKED ISSUE #320: no mo' spaces
We allow spaces in names. Maybe we shouldn't do that, it makes things look super weird.

</opinion>

rebeccaskinner:
fixes #320

### DIFF
--- a/parse/node.go
+++ b/parse/node.go
@@ -89,22 +89,17 @@ func (n *Node) Validate() error {
 }
 
 func validateName(name string) error {
-	var hasSpaces bool
-	var hasNonAlphaNum bool
 	var invalidRunes []rune
 	for _, letter := range []rune(name) {
 		if unicode.IsSpace(letter) {
-			hasSpaces = true
+			return errors.New("resource name may not contain spaces")
 		}
 		if !(unicode.IsLetter(letter) || unicode.IsNumber(letter) || letter == '-' || letter == '_' || letter == '.') {
-			hasNonAlphaNum = true
 			invalidRunes = append(invalidRunes, letter)
 		}
 	}
-	if hasSpaces {
-		return errors.New("resource name may not contain spaces")
-	}
-	if hasNonAlphaNum {
+
+	if invalidRunes != nil {
 		var badChars []string
 		printed := make(map[rune]struct{})
 		for _, runeChar := range invalidRunes {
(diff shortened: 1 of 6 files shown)


parse/node.go
```diff
@@ -81,10 +82,42 @@ func (n *Node) Validate() error {
...
+		if unicode.IsSpace(letter) {
+			hasSpaces = true
+		}
+		if !(unicode.IsLetter(letter) || unicode.IsNumber(letter) || letter == '-' || letter == '_' || letter == '.') {
+			hasNonAlphaNum = true
+			invalidRunes = append(invalidRunes, letter)
+		}
+	}
+	if hasSpaces {
+		return errors.New("resource name may not contain spaces")
```
BrianHicks: why not return this immediately instead of keeping track of another value and parsing the entire name?

parse/node.go
```diff
@@ -81,10 +82,42 @@ func (n *Node) Validate() error {
...
+func validateName(name string) error {
+	var hasSpaces bool
+	var hasNonAlphaNum bool
+	var invalidRunes []rune
+	for _, letter := range []rune(name) {
+		if unicode.IsSpace(letter) {
+			hasSpaces = true
+		}
+		if !(unicode.IsLetter(letter) || unicode.IsNumber(letter) || letter == '-' || letter == '_' || letter == '.') {
+			hasNonAlphaNum = true
```
BrianHicks: this isn't necessary, since you can just check if `invalidRunes == nil` later.

parse/node.go
```diff
@@ -81,10 +82,42 @@ func (n *Node) Validate() error {
...
+	if hasNonAlphaNum {
+		var badChars []string
+		printed := make(map[rune]struct{})
+		for _, runeChar := range invalidRunes {
+			if _, ok := printed[runeChar]; !ok {
+				badChars = append(badChars, string(runeChar))
+				printed[runeChar] = struct{}{}
+			}
+		}
+		return fmt.Errorf("invalid character(s) in resource name: %v", badChars)
```
BrianHicks: I assume this is a user-facing error message. Should it include something like "valid characters are unicode letters and numbers, dashes, underscores, and dots"?

BrianHicks:
LGTM! If you don't mind, though, I'm going to add some fuzzing for this before we merge it.

BrianHicks:
I'm going to add some fuzzing to this, too!

PR #537, opened 2016-12-06
Timer status display

### LINKED ISSUE #269: better status display on plan and execution
We have the following constraints:
- provide reasonable status to the operator to let them know the program has not frozen
- remove that status from the logs. Logs should only have actionable information!

My proposal: dynamically write a list to stdout. The list is the time the currently executing modules have been running and which are executing. Once the operation is complete, the list is erased. It updates every second otherwise. Something like:

```
root/module.a/task.do-the-thing: 10s
root/module.b/task.install-docker: 52s
```

Let the bikeshedding commence!

BrianHicks:
This adds an incremental status display for long-running tasks.

Fixes #269

### DIFF
--- a/cmd/plan.go
+++ b/cmd/plan.go
@@ -113,14 +113,15 @@ can be done separately to see what needs to be changed before execution.`,
 						"run":   resp.Run,
 						"id":    resp.Meta.Id,
 					})
-					if resp.Run == pb.StatusResponse_STARTED {
+					switch resp.Run {
+					case pb.StatusResponse_STARTED:
 						timer.AddTimer(resp.Meta.Id + ": " + resp.Stage.String())
 						slog.Info("got status")
-					} else {
-						slog.Debug("got status")
-					}
 
-					if resp.Run == pb.StatusResponse_FINISHED {
+					case pb.StatusResponse_FINISHED:
+						timer.RemoveTimer(resp.Meta.Id + ": " + resp.Stage.String())
+						slog.Debug("got status")
+
 						details := resp.GetDetails()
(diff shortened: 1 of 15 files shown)


cmd/timerdisplay.go
```diff
@@ -0,0 +1,143 @@
...
+	td.logger.Logger.Out = td.writer.Bypass()
+
+	td.wlock = new(sync.RWMutex)
+}
+
+// Message creates the current message to display
+func (td *TimerDisplay) Message() string {
+	td.once.Do(td.init)
+
+	out := "\nActive:\n"
```
arichardet: I noticed this output is sometimes displayed on one line and sometimes displayed twice, with the second a few lines below. The second instance flashes on and off the screen.

BrianHicks: Yes, that's a race condition in the library we're using. I'm not sure what's causing it, but it seems to be an issue with multi-line displays. Unfortunately, it looks awful on one line so I think we're just going to have to deal with it :\

cmd/plan.go
```diff
@@ -109,6 +114,7 @@ can be done separately to see what needs to be changed before execution.`,
 						"id":    resp.Meta.Id,
 					})
 					if resp.Run == pb.StatusResponse_STARTED {
+						timer.AddTimer(resp.Meta.Id + ": " + resp.Stage.String())
 						slog.Info("got status")
 					} else {
 						slog.Debug("got status")
```
BrianHicks: I actually forgot to change it back to Debug... I was just debugging.

PR #541, opened 2016-12-12
return a better error for planned network failures

### LINKED ISSUE #529: Return error instead of message from docker resource plan
Instead of returning a message with the error, an error should be returned.

Sample output:
```
root/docker.network.test-network:
 Error: resource cannot change because of an error
 Messages:
  gateway 192.168.129.1 already in use
 Has Changes: yes
 Changes:
  test2: "absent" => "present"

Summary: 2 errors, 2 changes

 * root/docker.network.test-network: resource cannot change because of an error
 * root: error in dependency "root/docker.network.test-network"
```

rebeccaskinner:
fixes #529

### DIFF
--- a/resource/docker/network/network.go
+++ b/resource/docker/network/network.go
@@ -79,15 +79,15 @@ func (n *Network) Check(context.Context, resource.Renderer) (resource.TaskStatus
 			n.diffNetwork(nw)
 		}
 
-		ok, err := n.isCreatable(nw)
+		invalidGWs, err := n.isCreatable(nw)
 		if err != nil {
 			n.RaiseLevel(resource.StatusFatal)
 			return n, err
 		}
 
-		if !ok {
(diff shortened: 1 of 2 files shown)


resource/docker/network/network.go
```diff
@@ -79,15 +79,15 @@ func (n *Network) Check(context.Context, resource.Renderer) (resource.TaskStatus
...
 		if err != nil {
 			n.RaiseLevel(resource.StatusFatal)
 			return n, err
 		}
 
-		if !ok {
+		if invalidGWs != nil && len(invalidGWs) > 0 {
 			n.RaiseLevel(resource.StatusCantChange)
-			return n, err
+			return n, fmt.Errorf("%v: gateway(s) are already in use", invalidGWs)
```
arichardet: When there is more than one gateway, readability is an issue. Maybe use some formatting to separate the values.

PR #543, opened 2016-12-13
Feature/file.fetch

arichardet:
Add file.fetch module:
- provide file path or url as source
- fetch with or without checksum
- includes a "force" option to replace the file if is already exists


resource/file/fetch/fetch.go
```diff
@@ -0,0 +1,217 @@
...
+
+// Check if the the file is on disk, and the hashes are availabe
+func (f *Fetch) Check(context.Context, resource.Renderer) (resource.TaskStatus, error) {
+	var (
+		hsh    hash.Hash
+		err    error
+		status = resource.NewStatus()
+	)
+
+	if !f.hasApplied {
```
rebeccaskinner: Could you do `if f.hasApplied { return status, nil }` here instead to reduce nesting of the rest of the code?

resource/file/fetch/fetch.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			status.RaiseLevel(resource.StatusCantChange)
+			return status, errors.Wrap(err, "will not attempt file fetch")
+		}
+	}
+
+	stat, err := f.DiffFile(status, hsh)
+	if err != nil {
+		return stat, err
+	} else if resource.AnyChanges(stat.Differences) {
```
rebeccaskinner: Would it be an option here to do `else if !resource.AnyChanges(state.Differences) { return status, nil }`?

resource/file/fetch/fetch.go
```diff
@@ -0,0 +1,217 @@
...
+			Mode: getter.ClientModeFile,
+		}
+		if err := client.Get(); err != nil {
+			status.RaiseLevel(resource.StatusFatal)
+			return status, errors.Wrap(err, "failed to fetch")
+		}
+		status.AddMessage("fetched successfully")
+		f.hasApplied = true
+	}
+
```
rebeccaskinner: I wonder if it might be worth it here to actually set `f.Hash` even if the user hadn't originally specified one- the idea being that someone might want to do a lookup on the hash of a fetched file even if they didn't actually specify an explicit one.  What do you think?

resource/file/fetch/preparer.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	if p.HashType != nil {
+		if !isValidHashType(*p.HashType) {
+			return nil, fmt.Errorf("\"hash_type\" must be one of \"%s,%s,%s,%s\"", string(HashMD5), string(HashSHA1), string(HashSHA256), string(HashSHA512))
+		}
+	}
+
+	if p.Hash != nil {
+		if strings.TrimSpace(*p.Hash) == "" {
```
rebeccaskinner: Is it worth actually checking the length of the string here to make sure it's the right length for a hash of the specified type?

rebeccaskinner:
LGTM after vendor fixing and rebasing on #533

PR #549, opened 2016-12-19
Feature/74 file owner

### LINKED ISSUE #74: builtin file owner module

rebeccaskinner:
fixes #74


resource/file/owner/OSExecutor.go
```diff
@@ -0,0 +1,92 @@
...
+func (o *OSExecutor) Walk(root string, f filepath.WalkFunc) error {
+	return filepath.Walk(root, f)
+}
+
+// Chown returns the result of OS.Chown
+func (o *OSExecutor) Chown(path string, uid, gid int) error {
+	return os.Chown(path, uid, gid)
+}
+
+func osStat(path string) (*syscall.Stat_t, error) {
```
arichardet: maybe move this to the end of the file?

resource/file/owner/owner_test.go
```diff
@@ -0,0 +1,354 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestInterface ensures that owner.Owner implements the resource.Task interface
+func TestInterface(t *testing.T) {
+	t.Parallel()
+	assert.Implements(t, (*resource.Task)(nil), new(owner.Owner))
+}
+
```
arichardet: maybe add some tests to check/apply when destination is and is not a directory with recursive = true

resource/file/owner/owner.go
```diff
@@ -94,20 +118,30 @@ func (o *Owner) Check(context.Context, resource.Renderer) (resource.TaskStatus,
...
 }
 
 // Apply applies the ownership to the file
 func (o *Owner) Apply(context.Context) (resource.TaskStatus, error) {
-	showDetails := !(o.Recursive && o.HideDetails)
 	status := resource.NewStatus()
+	showDetails := !(o.Recursive && o.HideDetails)
+
+	if _, err := os.Stat(o.Destination); os.IsNotExist(err) {
+		return nil, fmt.Errorf("cannot change ownership of non-existant file: %s", o.Destination)
```
arichardet: correct spelling is existent

resource/file/owner/preparer.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if nil != p.UID {
+		uidStr = strconv.Itoa(*p.UID)
+	}
+
+	if nil != p.GID {
+		gidStr = strconv.Itoa(*p.GID)
+	}
+
```
rebeccaskinner: We shouldn't need to do an explicit check here because we're looking up the username and group name based on the uid/gid and will get an error back if the value is out of range.

PR #551, opened 2016-12-19
when apply fails due to changes still being present; use those diffs …

rebeccaskinner:
…for the status


BrianHicks:
Lettuce Goats Together Monday

PR #553, opened 2016-12-19
Better Makefiles

BrianHicks:
Improves our makefiles to be faster and stronger. You can build docs completely in one step now!


rebeccaskinner:
LGTM on green

PR #554, opened 2016-12-19
Fix Linting Bugs

BrianHicks:
This fixes all the outstanding linting bugs, except those in aligncheck (which I've removed; it wasn't getting us much benefit.)


rebeccaskinner:
LGTM on green

PR #557, opened 2016-12-21
reset updates needed after apply in lvm

rebeccaskinner:
This doesn't have unit tests because of the complexity of the existing test mock infrastructure, but is covered by the lvm sample.


arichardet:
LGTM!

PR #570, opened 2016-12-28
Sleep in Graph Generation Test

BrianHicks:
There wasn't any sleeping in the graph generation test. I'm not sure why it didn't give us problems before, but it is now. This PR adds a quarter second sleep between starting the server and using it.

### DIFF
--- a/blackbox/test_graph_generation.sh
+++ b/blackbox/test_graph_generation.sh
@@ -6,7 +6,7 @@ TMP=$(mktemp -d -t converge.graphviz.XXXXXXXXXX)
 
 "$ROOT"/converge server --no-token &
 PID=$!
-sleep 1
+sleep 0.25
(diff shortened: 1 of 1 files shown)


blackbox/test_graph_generation.sh
```diff
@@ -6,6 +6,7 @@ TMP=$(mktemp -d -t converge.graphviz.XXXXXXXXXX)
 
 "$ROOT"/converge server --no-token &
 PID=$!
+sleep 0.25
```
rebeccaskinner: Could you change this to 1 second?  Testing in the CI container locally seems to indicate than anything less that a second doesn't reliably fix the problem.

BrianHicks:
@rebeccaskinner increased. It passes, etc.

PR #571, opened 2016-12-28
Add various generators

### LINKED ISSUE #560: Autocompletion and man pages
Cobra will automatically generate man pages and bash autocompletion for us. We should add that. I like Cockroach's approach:

```console
$ cockroach gen -h
Generate manpages, example shell settings, example databases, etc.

Usage:
  cockroach gen [command] [flags]
  cockroach gen [command]

Available Commands:
  man          generate man pages for CockroachDB
  autocomplete generate bash autocompletion script for CockroachDB
```

BrianHicks:
These are provided for us since we're using cobra, this just enables them.

- `converge gen autocomplete` generates bash autocompletion
- `converge gen man` generates man pages. I've chosen section 8, "System Administration tools and daemons"
- `converge gen docs` generates a command reference. I hooked this into our docs so we get command reference docs.

Fixes #560


cmd/gen.go
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import "github.com/spf13/cobra"
+
+// genCmd represents the gen command
+var genCmd = &cobra.Command{
+	Use: "gen",
```
arichardet: Please add a generic description for `gen`

docs/content/resources/docker.container.md
```diff
@@ -1,197 +0,0 @@
----
-title: "docker.container"
```
arichardet: All of the docs are removed. Running `make` in /docs fails with `unknown command "gen" for "converge"`, but using the new `gen docs` command places docs in /markdown-docs.

arichardet:
Current thing to change as we have discussed: updating how the man page is generated/where it is placed (i.e. man1/man8 and the file names) so they can be displayed.

A general comment:
It would be good to add instructions or point the user to a resource where they can get information on the rest of the setup for converge man, like adding the MANPATH environment variable, etc. as well as the required setup for bash autocomplete. I think most will assume after they have completed the setup for man, running `man converge` would output man pages, so having this documented would be very helpful.

docs/content/resources/docker.container.md
BrianHicks: This works for me. Do you have binaries laying around?

docs/Makefile
```diff
@@ -2,7 +2,7 @@ all: public
 
 -include resources.mk
 
-public: content/*.md static/**/* ${RESOURCES} content/license.md static/images/**/*.png
+public: content/commands content/*.md static/**/* ${RESOURCES} content/license.md static/images/**/*.png
```
arichardet: it looks like everything in `static/images/**/*.png` is also taken care of with `static/**/*`
I know this is not a change you made for this PR, but something I noticed.

BrianHicks:
@arichardet ready for another review here

arichardet:
@BrianHicks please take a look at your failing test `blackbox/test_man.sh`

PR #572, opened 2016-12-28
Update docs style with new branding

BrianHicks:
We've got a style guide now. I've applied it to the docs. Hooray!

Also some minor tweaks:

- reversed the version numbers
- added a temporary style for code style background, I'll be writing an HCL parser so we can get proper highlighting soon.

image


arichardet:
LGTM!

PR #581, opened 2017-01-06
Feature/user expiry

arichardet:
Adds the ability to specify a user expiry when adding or modifying a user.

### DIFF
--- a/resource/user/preparer.go
+++ b/resource/user/preparer.go
@@ -17,6 +17,7 @@ package user
 import (
 	"fmt"
 	"math"
+	"time"
 
 	"github.com/asteris-llc/converge/load/registry"
 	"github.com/asteris-llc/converge/resource"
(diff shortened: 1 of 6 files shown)


resource/user/preparer.go
```diff
@@ -72,6 +73,12 @@ type Preparer struct {
 	// HomeDir must also be indicated if MoveDir is set to true.
 	MoveDir bool `hcl:"move_dir"`
 
+	// Expiry is the date on which the user account will be disabled. The date is
+	// specified in the format YYYY-MM-DD. If not specified, the default expiry
+	// date specified by the EXPIRE variable in /ect/default/useradd, or an empty
```
rebeccaskinner: Typo.  Should be `/etc/default/useradd`

resource/user/preparer.go
```diff
@@ -72,6 +73,12 @@ type Preparer struct {
...
+	// string (no expiry) will be used by default.
+	Expiry time.Time `hcl:"expiry"`
```
rebeccaskinner: Should this be `*time.Time`?  Otherwise any other changes to a user will have the side effect of setting an expiry since we'll always be setting it.

arichardet: I only set the expiry if it is specified in the hcl. There are checks in user.go (DiffAdd and DiffMod) for this.

Issue #445, opened 2016-10-28
Refactor User Delete to use Diffs


arichardet:
Move logic out of Check and Apply and into a diff function for deleting a user.

rebeccaskinner:
fixed in #584

PR #584, opened 2017-01-10
Feature/userdel diffs

arichardet:
Improve diffs used when deleting a user. 
Implements #445

### DIFF
--- a/resource/user/user.go
+++ b/resource/user/user.go
@@ -131,19 +131,10 @@ func NewUser(system SystemUtils) *User {
 
 // Check if a user user exists
 func (u *User) Check(context.Context, resource.Renderer) (resource.TaskStatus, error) {
-	var (
-		userByID *user.User
-		uidErr   error
-	)
-
-	// lookup the user by name and lookup the user by uid
-	// the lookups return ErrUnsupported if the system is not supported
+	// lookup the user by name
+	// ErrUnsupported is returned if the system is not supported
 	// Lookup returns user.UnknownUserError if the user is not found
-	// LookupID returns user.UnknownUserIdError if the uid is not found
 	userByName, nameErr := u.system.Lookup(u.Username)
-	if u.UID != "" {
-		userByID, uidErr = u.system.LookupID(u.UID)
-	}
 
 	status := resource.NewStatus()
 
@@ -152,10 +143,10 @@ func (u *User) Check(context.Context, resource.Renderer) (resource.TaskStatus, e
 		return status, ErrUnsupported
 	}
 
+	_, nameNotFound := nameErr.(user.UnknownUserError)
+
 	switch u.State {
 	case StatePresent:
-		_, nameNotFound := nameErr.(user.UnknownUserError)
-
 		switch {
 		case nameNotFound:
 			_, err := u.DiffAdd(status)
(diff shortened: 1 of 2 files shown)


resource/user/user.go
```diff
@@ -217,19 +183,10 @@ func (u *User) Check(context.Context, resource.Renderer) (resource.TaskStatus, e
 
 // Apply changes for user
 func (u *User) Apply(context.Context) (resource.TaskStatus, error) {
```
BrianHicks: why couldn't apply use the status that you've created already, again? I feel like we had this discussion, but a) I don't remember and b) has the answer changed?

resource/user/user.go
```diff
@@ -401,6 +336,40 @@ func (u *User) DiffAdd(status *resource.Status) (*AddUserOptions, error) {
 	return options, nil
 }
 
+// DiffDel checks for differences between the current and desired state for the
+// user to be deleted indicated by the User fields.
+func (u *User) DiffDel(status *resource.Status, userByName *user.User, nameNotFound bool) error {
```
BrianHicks: nitpick: should this be public?

resource/user/user.go
```diff
@@ -401,6 +336,40 @@ func (u *User) DiffAdd(status *resource.Status) (*AddUserOptions, error) {
...
+	switch {
+	case nameNotFound:
+		return nil
+	case userByName != nil:
+		switch {
```
BrianHicks: nitpick: this `if … else` in fancier clothes

resource/user/user.go
```diff
@@ -401,6 +336,40 @@ func (u *User) DiffAdd(status *resource.Status) (*AddUserOptions, error) {
...
+			case userByID != nil && *userByID != *userByName:
+				status.RaiseLevel(resource.StatusCantChange)
+				return fmt.Errorf("uid %s belongs to different user", u.UID)
+			case userByID != nil && *userByID == *userByName:
+				status.AddDifference("user", u.Username, fmt.Sprintf("<%s>", string(StateAbsent)), "")
+			}
+		}
+	}
+
+	if resource.AnyChanges(status.Differences) {
```
BrianHicks: possible enhancement: why isn't this a method on `status`? (cc @rebeccaskinner)

resource/user/user.go
arichardet: I don't believe we can since we no longer keep `status` as a field of the task struct.

resource/user/user.go
arichardet: It makes sense to be private, but is public for testing purposes.

BrianHicks: IMO it would be better to make the tests part of the package than it would be to expose extra members. But it's definitely not a blocker. :)

resource/user/user.go
arichardet: I thought you liked switch better. ;) I will change it back to `if...else`.

resource/user/user.go
arichardet: After some discussion, we've concluded to leave as is for `user` since testing is currently performed for a `package <name>` in `package <name_test>`. In the next resource added to  Converge, we will try testing as part of the package.

BrianHicks: :metal: ticket?

arichardet: Opened #585

Issue #513, opened 2016-11-21
unarchive resource


ryane:
consider adding a resource that makes it easier to extract files from tar, tar.gz, and zip files.

idea for a minimal implementation:

```hcl
file.unarchive "etcd" {
  source      = "/tmp/etcd.tar.gz" # or https://github.com/coreos/etcd/releases/download/v3.0.14/etcd-v3.0.14-linux-amd64.tar.gz
  destination = "/usr/local/bin/"
  owner       = "root"
  group       = "root"
}
```

rebeccaskinner:
closed in #592

automaticgiant:
any ideas to make it idempotent?

rebeccaskinner:
There's not a good way to ensure it's idempotent without keeping a registry of the archives and their checksums.  For a lot of formats, e.g. tar, you can get a list of files pretty easily, but you'd still ned to extract the archive to get checksums for each of the files; at that point there's not much value in calculating the checksum and only copying over the file if it differs.  We might be able to make check more useful but we'd need to have some way to deal with changes of remote resources between plan and apply.  We'd talked about having check generate a database that would be used for execution apply, if all resources were fetched at check time and serialized into that a database we could make some guarantees there.

automaticgiant:
i've rewritten my unarchive invocation of concern as:
```
task untgz_converge {
  check = "[ {{lookup `file.fetch.converge.destination`}} -ot {{param `bin`}}converge ]"
  apply = "tar xzm -C {{platform.OS}}/ -f {{platform.OS}}/.archives/converge.tgz converge"
}
```
thanks for the thoughts

PR #592, opened 2017-02-03
Feature/file.unarchive

arichardet:
Feature implementation of unarchive #513 
The decision was made to not include owner/group in unarchive, and instead allow the user to utilize the file.owner resource for this.


resource/unarchive/unarchive.go
```diff
@@ -0,0 +1,383 @@
...
+
+			faStat, err := os.Stat(fA)
+			if err != nil {
+				return err
+			}
+			fbStat, err := os.Stat(fB)
+			if err != nil {
+				return err
+			}
+
```
rebeccaskinner: Just a thought here, but would it make sense to compare the size of the files before doing a checksum?  It seems like in the most common case the files would differ in size as well as contents, so we could skip the actual time it takes to generate a checksum.

resource/unarchive/unarchive.go
```diff
@@ -0,0 +1,383 @@
...
+					return err
+				}
+
+				// get src FileInfo
+				srcInfo, err := src.Stat()
+				if err != nil {
+					return err
+				}
+
+				err = ioutil.WriteFile(u.destDir.Name()+fileName, srcData, srcInfo.Mode().Perm())
```
rebeccaskinner: This could exhaust the memory on the system if the file is larger than free ram; should we either shell out to `cp` or memory map https://godoc.org/golang.org/x/exp/mmap the file?

resource/unarchive/unarchive.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// setFetchLoc sets the location for the fetch destination
+func (u *Unarchive) setFetchLoc() {
+	if u.fetchLoc != "" {
+		return
+	}
+
+	_, file := filepath.Split(u.Source)
+	u.fetchLoc = "/var/run/converge/cache/" + file
```
rebeccaskinner: Maybe this should be some sort of var that we can change/set easily, especially if we decide to move it around based on the UID of the user running converge to avoid permission problems.

resource/unarchive/preparer.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// Preparer for Unarchive
+//
+// Unarchive renders unarchive data
+type Preparer struct {
+	// Source to unarchive - must exist locally
+	Source string `hcl:"source" required:"true" nonempty:"true"`
+
+	// Destination for the unarchive
```
rebeccaskinner: Should the docstring here specify that the destination should be a directory?

PR #604, opened 2017-02-24
Feature/error resource

rebeccaskinner:


BrianHicks:
We're not checking in the generated resource content anymore, could you remove it?

PR #610, opened 2017-03-01
docs: highlight resource reference, minor edits

arichardet:
Docs changes based on user feedback

### DIFF
--- a/docs/content/commands.md
+++ b/docs/content/commands.md
@@ -9,9 +9,10 @@ menu:
 
 ---
 
-Converge has resources that you can use to make various changes on your system.
-For example, `file.content` will control the content of a file, and `file.mode`
-will control that file's mode.
+Converge is a tool you can use to make various changes on your system. Commands
+such as `plan` and `apply` allow you to plan and apply changes to your system,
+respectively. Use the `graph` command to visualize the execution of your
+changes.
 
-For more details on how to use the resources, see the
-[getting started guide]({{< ref "getting-started.md" >}}).
+Each command has a dedicated section detailing its use, options, and a listing
+of related commands.
--- a/docs/content/commands.md
+++ b/docs/content/commands.md
@@ -9,8 +9,8 @@ menu:
 
 ---
 
-Converge is a tool you can use to make various changes on your system. Commands
-such as `plan` and `apply` allow you to plan and apply changes to your system,
+Use Converge to be sure your system is set up the way you like. `plan` and
+`apply` commands allow you to plan and apply changes to your system,
 respectively. Use the `graph` command to visualize the execution of your
 changes.
 
--- a/docs/content/dependencies.md
+++ b/docs/content/dependencies.md
@@ -18,24 +18,23 @@ are!
 Briefly, Converge operates by thinking about your deployment as a graph of tasks
 that need to be done. It walks around the graph from the leaves (AKA tasks with
 no dependencies) all the way to to the root (AKA Rome, where all roads lead.)
-Let's explore what that means with one of our graphs from before:
+Let's explore what that means with a graph from the [params example in the
+getting started]({{< ref "getting-started.md" >}}#params) guide:
 
 {{< figure src="/images/getting-started/hello-world-params.png"
            caption="A graph with a parameter. The file hello.txt depends on the name parameter." >}}
 
-What does Converge do when you ask it to apply this graph? When Converge loads
-this file, it will load then file and then start walking at the node that
-doesn't have any dependencies. In this case, that's `param.name = "World"`. When
-`param.name` has been successfully walked, it will move on to `File: hello.txt`.
-If we're successful, the root (`/`) will be marked as successful, and our graph
-will be successful. Neat!
+What does Converge do when you ask it to apply this graph? Converge will load
+the file and start walking at a node that has no dependencies. In this case,
+that's `param.name = "World"`. When `param.name` has been successfully walked,
+Converge will move on to `File: hello.txt`, and then to the root (`/`). Once
+each node is marked as successful, our graph walk is complete. Neat!
 
 ## The Graph Command
 
-All the graphs we've been seeing so far have just been the output of Converge's
-`graph` command. When asked, Converge will load up any modules you specify and
-then render them as [Graphviz](http://graphviz.org/) dot output. You can render
-that like so:
+All the graphs we've seen so far have just been the output of Converge's `graph`
+command. When asked, Converge will load up any modules you specify and render
+them as [Graphviz](http://graphviz.org/) dot output. You can render like so:
 
 ```bash
 $ converge graph --local yourModule.hcl | dot -Tpng > yourModule.png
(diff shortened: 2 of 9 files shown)


docs/content/commands.md
```diff
@@ -9,9 +9,10 @@ menu:
 
 ---
 
-Converge has resources that you can use to make various changes on your system.
-For example, `file.content` will control the content of a file, and `file.mode`
-will control that file's mode.
+Converge is a tool you can use to make various changes on your system. Commands
```
BrianHicks: How about

> You can use Converge to make sure your system is set up the way you like.

docs/content/commands.md
```diff
@@ -9,9 +9,10 @@ menu:
...
+such as `plan` and `apply` allow you to plan and apply changes to your system,
```
BrianHicks: You can remove "Commands such as" from this to make it stronger

docs/content/dependencies.md
```diff
@@ -18,24 +18,23 @@ are!
...
-this file, it will load then file and then start walking at the node that
-doesn't have any dependencies. In this case, that's `param.name = "World"`. When
-`param.name` has been successfully walked, it will move on to `File: hello.txt`.
-If we're successful, the root (`/`) will be marked as successful, and our graph
-will be successful. Neat!
+What does Converge do when you ask it to apply this graph? Converge will load
+the file and start walking at a node that has no dependencies. In this case,
+that's `param.name = "World"`. When `param.name` has been successfully walked,
+Converge will move on to `File: hello.txt`, and then to the root (`/`). Once
+each node is marked as successful, our graph walk is complete. Neat!
```
BrianHicks: now that I'm looking at this again, it's talking about implementation details. The user's unlikely to care about how params are walked. What would you think about rewriting this with two resources (the user certainly cares how two *resources* are walked) or removing?

docs/content/dependencies.md
```diff
@@ -18,24 +18,23 @@ are!
...
 
 ## The Graph Command
 
-All the graphs we've been seeing so far have just been the output of Converge's
-`graph` command. When asked, Converge will load up any modules you specify and
-then render them as [Graphviz](http://graphviz.org/) dot output. You can render
-that like so:
+All the graphs we've seen so far have just been the output of Converge's `graph`
+command. When asked, Converge will load up any modules you specify and render
+them as [Graphviz](http://graphviz.org/) dot output. You can render like so:
```
BrianHicks: how about "You can render your own modules like this:"?

docs/content/dependencies.md
```diff
@@ -46,13 +45,12 @@ makes it easier to think about how the graph will be executed.
...
-Resources may references one-another as long as the references do not introduce
-circular dependencies.  When creating a reference from one node to another we
-can use the `lookup` command to reference fields of an entry that are provided
-by that entries module.  The available fields will vary depending on the module
-and should be documented along with each module.  The example below illustrates
-using `lookup` to access fields from a `docker.image` node from within
-`docker.container`
+Resources may reference one another as long as the references do not introduce
+circular dependencies. When creating a reference from one node to another, we
+can use the `lookup` command to reference fields of another entry's module. The
```
BrianHicks: how about "to reference fields from another resource"?

docs/content/dependencies.md
```diff
@@ -46,13 +45,12 @@ makes it easier to think about how the graph will be executed.
...
+available fields will vary depending on the module and should be documented
+along with each module. The example below illustrates using `lookup` to access
```
BrianHicks: this sentence ("will vary depending on the module") could just be replaced with a link to the module docs root and a pointer to look there for the specific fields. As is, this is kind of obvious but unhelpful.

BrianHicks: Also, how about "The example below uses `lookup` to get the right tag from a Docker image"?

docs/content/dependencies.md
```diff
@@ -112,8 +110,8 @@ how it looks:
 
 To fix this, we'll need to specify `depends` on our `file.content`. `depends` is
 a list of resources in the current module that must be successfully walked
-before walking ours. They're specified as the resource type, a dot, then the
-resource name. So `task "names"` above becomes `task.names`.
+before walking the containing resource. They're specified as the resource type,
```
BrianHicks: Why don't we remove references to walking here. I could see them confusing new users. "… that must be successful before beginning the current resource" could work?

docs/content/getting-started.md
```diff
@@ -181,10 +183,10 @@ this:
 
 ## Modules Calling Modules
 
-This is all well and good, but we don't want to have to write things the same
-*every time* right? Well, good news: that's what modules are for! Now that
-you've written a module, you can require it from any other module to add it to
-your tree. Create a new module, let's call it `helloYou.hcl`:
+Let's look at how we can take advantage of reusability. Good news: that's what
+modules are for! Now that you've written a module, you can require it from any
+other module to add it to your tree. Create a new module, let's call it
```
BrianHicks: We could lose "Create a new module, let's call it `helloWorld.hcl`." in favor of the imperative sentence following it.

docs/content/index.md
```diff
@@ -15,6 +15,9 @@ laptops and other devices.
 Key features:
 
 - Easy to install and run. A single binary and configuration file is all you need.
+- Resources to make changes on your system: docker images and containers, users
```
BrianHicks: Could you add these changes to `README` too, please?

docs/content/index.md
```diff
@@ -54,9 +57,9 @@ file.content "render" {
 }
 ```
 
-Invoke this with `converge apply --local samples/fileContent.hcl` to install
-Traefik from yum on your system. You can also `converge plan --local
-samples/fileContent.hcl` to see what changes will be made before you apply them.
+It is a good idea to `converge plan --local samples/fileContent.hcl` to see what
+changes will be made before you apply them. Invoke this with `converge apply
```
BrianHicks: How about "If you like what you see, you can install Traefik on your system by running `converge apply …`."?

docs/content/resources.md
```diff
@@ -9,9 +9,12 @@ menu:
 
 ---
 
-Converge has resources that you can use to make various changes on your system.
-For example, `file.content` will control the content of a file, and `file.mode`
-will control that file's mode.
+Converge has resources you can use to make various changes on your system. For
+example, `file.content` will control the content of a file, and `file.mode`
+will control a file's mode.
```
BrianHicks: What about saying "`file.content` controls what's in a file, and `file.mode` sets its permissions." instead of repeating "content" and "mode"?

docs/content/dependencies.md
arichardet: I lean more toward taking this out. You were referring to entire Graph Walking section, right?

docs/content/getting-started.md
arichardet: I see what you are saying. I was trying to be explicit b/c I think there could be confusion with how this relates to the previous example and that this new module is called `helloWorld.hcl` within a file named `helloYou.hcl`. I changed it up.

docs/content/dependencies.md
BrianHicks: Well, I don't know… knowing how we walk the graph is useful information. Buuuut you almost never need to reason about it terms of param dependencies, only module and resources. So I'd say changing the example is the better option.

arichardet:
@BrianHicks, this is ready for another look

PR #616, opened 2017-03-03
systemd docs: fix links to systemd properties

arichardet:
fix broken links to systemd properties in docs


BrianHicks:
Let's Go Team Marshmallow!

PR #633, opened 2017-09-07
Be sure that script is fully downloaded before run

hauleth:
This also removes some smells there and here like needless two step
checks for command status. However there is still much room to for
improvements.

### DIFF
--- a/install-converge.sh
+++ b/install-converge.sh
@@ -24,227 +24,217 @@
 
 set -e
 
-# Default to latest stable
-version="0.6.0"
+__run() {
+	# Default to latest stable
+	version="0.6.0"
 
-install_dir="/usr/local/bin"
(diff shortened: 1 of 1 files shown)


hauleth:
As there is a lot of whitespace changes it is the best to view it with [whitespace changes ignored](https://github.com/asteris-llc/converge/pull/633/files?w=1) (you can add `?w=1` at the end of URL to achieve that).

rebeccaskinner:
Thanks for the PR; I'll get that merged soon.  We have some wercker issues that make it difficult to merge directly from fork forked repos, but it should go in this week.

hauleth:
@rebeccaskinner the point is that this PR doesn't touch any of the Go files and `./install-converge.sh` isn't covered in test suite from what I seen.

PR #636, opened 2017-11-15
Convert from glide to go dep

rebeccaskinner:
this fix changes the vendoring tool from glide to go dep, and includes a bugfix that had been pending as well as a makefile fix.


stevendborrelli:
For some reason, wercker did not work on this, I have been able to test successfully.

